PR #69, opened 2023-05-16
WiP: proposal for shortcuts

jeanregisser:
### Description

This is a proposal for adding support for shortcut hooks.

### Details

This is focusing on applying shortcuts to positions. But it could be extended to support shortcuts for tokens or more generic actions.

It would support the following use cases
- claiming Good Dollar UBI
- claiming Ubeswap farm staking rewards

The shortcuts would be defined in a new file under `src/apps/{appId}/shortcuts.ts`. This is to give builders clear visibility into what type of hooks are supported for a given app.
The plan is to rename `plugin.ts` to `positions.ts` to follow with this new structure.

The positions are extended with a new `availableShortcutIds` property which allows defining which shortcuts can be applied to them. Underlying tokens get a new optional `category` property which allows giving more meaning to what they are.
Specifically here, the `claimable` category  tells the wallet the tokens and amounts which can be claimed.

The wallet would fetch the shortcuts with a new route in `hooks-api`.
(Or we could also return both positions and shortcuts data in a single call).

Using the new data in the returned positions, the wallet would be able to fill the content of a new "Claim Rewards" screen.
Displaying details for the claimable tokens.
Tapping the button next to each of them would call the shortcut's `onTrigger` function, via a new route in `hooks-api`.
It would return 0, 1 or more transaction to sign.

The wallet would take it from there and display the usual transaction signing bottom sheet. Similar to what's happening when interacting with dapps with WalletConnect.

### Future considerations

This could be extended to support other types of shortcuts related to positions and requiring additional input parameters. For instance adding/removing liquidity in Ubeswap pools. This would require a new wallet UI to specify the amounts needed.

We could maybe think of pre-built UI elements that could be used and defined by shortcuts so we don't have to build custom UIs in the wallet to support them.

### Related linear task

Fixes RET-732


src/shortcuts.ts
```diff
@@ -0,0 +1,19 @@
+export interface ShortcutsHook {
+  getShortcutDefinitions(): ShortcutDefinition[]
+}
+
+export interface ShortcutDefinition {
```
silasbw, line 5: Did you consider adding a `getText` method?  I'm wondering if that would meet the requirements you're aiming for below (rather than making a confident suggestion on a change).

> The positions are extended with a new availableShortcutIds property which allows defining which shortcuts can be applied to them. Underlying tokens get a new optional category property which allows giving more meaning to what they are.
Specifically here, the claimable category tells the wallet the tokens and amounts which can be claimed.

src/shortcuts.ts
```diff
@@ -0,0 +1,19 @@
...
+  id: string // Example: claim-reward
+  name: string // Example: Claim
+  description: string // Example: Claim your reward
+  networks: string[] // Example: ['celo']
+  category?: 'claim' // We'll add more categories later
+  onTrigger: (
```
dievazqu, line 11: Probably this `onTrigger` can receive an extra parameter to allow shortcuts with user inputs?
But I guess we can add it later.

src/apps/ubeswap/plugin.ts
```diff
@@ -205,8 +205,10 @@ async function getFarmPositionDefinitions(
           {
             address: farm.rewardsTokenAddress.toLowerCase(),
             network,
+            category: 'claimable',
           },
         ],
+        availableShortcutIds: ['claim-reward'],
```
dievazqu, line 211: It seems to me, that each shortcut might want to send different information to the client.
For example, the `claim-reward` probably wants to indicate the rewarded token and amount. I'm not sure if adding the `category: 'claimable',` is the best abstraction to achieve this.
What would happen if the reward is in another token? Or what if the amounts are changing related to an external factor (like the date).

So, I think this list should be something more like:

```
availableShortcuts: [{
id: 'claim-reward',
params: {
 rewardedToken: '0x1234',
 rewardedAmount: 1
}],
```

But I haven't thought about this too much, so maybe your approach is considering something that I'm missing.

dievazqu: Another option would be to send this info when fetching the shortcuts instead of the positions, so if there is a huge calculation that we need to do to show some shortcut data to the user, we are not blocking the positions with it.

src/shortcuts.ts
jeanregisser: Thanks for the suggestion, one drawback of adding a `getText` method is it puts more constraints for the display done by the wallet. The wallet already has good components for displaying amounts correctly with the right local currency and decimal symbol (based on the user preferences). Also what if we wanted to make one part of the string bold or something?
We'd need to pass more user context here to be able to do that. Or find a way to represent that with a custom syntax.

For now I preferred letting the wallet use the full position object and decide how it wants to display it.
I had to add the `category` type though to the tokens so the wallet can tell which part is the reward. 
But I see other benefits of this for future features. We could for instance use this info to create a breakdown of value: total claimable, total debt, total collateral, etc
Or if we build a detail screen for positions, we could add a visual difference to the claimable tokens (+ an action button there directly for the shortcut).

src/shortcuts.ts
jeanregisser: Yes it would need to be extended for uses cases with user inputs.

src/apps/ubeswap/plugin.ts
jeanregisser: > It seems to me, that each shortcut might want to send different information to the client. For example, the `claim-reward` probably wants to indicate the rewarded token and amount. I'm not sure if adding the `category: 'claimable',` is the best abstraction to achieve this. What would happen if the reward is in another token? Or what if the amounts are changing related to an external factor (like the date).
> 
> So, I think this list should be something more like:
> 
> ```
> availableShortcuts: [{
> id: 'claim-reward',
> params: {
>  rewardedToken: '0x1234',
>  rewardedAmount: 1
> }],
> ```
> 
> But I haven't thought about this too much, so maybe your approach is considering something that I'm missing.

Yes this is why I added the `category: 'claimable'` to the tokens.
I explained this more in this comment I just made: https://github.com/valora-inc/hooks/pull/69#discussion_r1196255089

What you're suggesting would also work. I represented it a bit differently with other possible future use cases in mind.

jeanregisser: > Another option would be to send this info when fetching the shortcuts instead of the positions, so if there is a huge calculation that we need to do to show some shortcut data to the user, we are not blocking the positions with it.

I see, I think we're fine expecting fetching the positions is where all the hard work happen. All amounts should be returned by positions. And this is one of the important piece of data needed for shortcuts.

I wanted to avoid redoing requests / calculations to fetch shortcuts (definitions).
That's why I ended up "tagging" positions with very light shortcut specific info.
After all, the current shortcuts we need are related to positions. We could still extend that to things that aren't positions though.

I also tried to see what it would look like to have shortcuts use the output of positions. But then it needed to dig through the positions and figure out which part of the positions are the rewards. It didn't look right.

silasbw (approved):

PR #73, opened 2023-05-23
fix(ubeswap): use static list of farms

jeanregisser:
This is a temporary solution to unblock the work on hooks, since fetching `FarmInfo` events now times out with Forno.

- farms info is now stored in a static json file
- a script was added to generate it

Farms don't change very often anyway.

This also speeds up querying Ubeswap positions. From more than 20-30 secs down to ~5 secs.

One medium/long term solution would be to index farms using TheGraph.
I've started exploring this, but the initial syncing with The Graph takes a long time and makes it a bit hard to iterate on this.
https://thegraph.com/hosted-service/subgraph/jeanregisser/ubeswap?selected=logs

See Slack [thread](https://valora-app.slack.com/archives/C04NFRMSZQX/p1684242639374869).

Fixes RET-724


src/apps/ubeswap/scripts/updateFarms.ts
```diff
@@ -0,0 +1,57 @@
...
+import { FarmInfoEventAbi } from '../abis/farm-registry'
+import { createPublicClient, http } from 'viem'
+import { celo } from 'viem/chains'
+
+const FARM_REGISTRY = '0xa2bf67e12EeEDA23C7cA1e5a34ae2441a17789Ec'
+const FARM_CREATION_BLOCK = 9840049n
+
+// How many blocks to query at once
+// Decrease this if the RPC node times out
+const BLOCKS_PER_QUERY = 500_000n
```
silasbw, line 16: fwiw, I think this is considered a pretty big chunk.  quicknode, for example, only allows 10_000 block sized chunks.

silasbw (approved):

PR #74, opened 2023-05-24
chore: add image urls for existing positions

kathaypacific:
For now we'll use the dapp logo as the position image. This PR adds the images from our dapps list for halo-fi and ubeswap, and the Celo assets image for locked celo.

Fixes RET-734


jeanregisser (approved):
👍

PR #100, opened 2023-06-22
feat: Moola plugin

dievazqu:
### Description

Adding a plugin for Moola: debt tokens are mapped to a position with the negative `usdPrice` of the base token.

### DIFF
--- a/src/apps/moola/debtTokens.ts
+++ b/src/apps/moola/debtTokens.ts
@@ -13,7 +13,7 @@ export const MOOLA_DEBT_TOKENS: DebtTokenDefinition[] = [
     baseTokenAddress: '0xd8763cba276a3738e6de85b4b3bf5fded6d6ca73',
     debtTokenAddress: '0xfb6c830c13d8322b31b282ef1fe85cbb669d9ae8',
     imageUrl:
-      'https://raw.githubusercontent.com/valora-inc/address-metadata/main/assets/tokens/mcEUR.png',
+      'https://raw.githubusercontent.com/valora-inc/dapp-list/main/assets/moola.png',
     description: 'Moola variable debt',
     title: 'cEUR debt',
   },
@@ -21,7 +21,7 @@ export const MOOLA_DEBT_TOKENS: DebtTokenDefinition[] = [
     baseTokenAddress: '0x765de816845861e75a25fca122bb6898b8b1282a',
     debtTokenAddress: '0xf602d9617564c07f1e128687798d8c699ced3961',
     imageUrl:
-      'https://raw.githubusercontent.com/valora-inc/address-metadata/main/assets/tokens/mcUSD.png',
+      'https://raw.githubusercontent.com/valora-inc/dapp-list/main/assets/moola.png',
     description: 'Moola variable debt',
     title: 'cUSD debt',
   },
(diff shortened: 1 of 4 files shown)


src/apps/moola/debtTokens.ts
```diff
@@ -0,0 +1,76 @@
...
+    title: string
+    description: string
+  }
+  
+export const MOOLA_DEBT_TOKENS: DebtTokenDefinition[] = [
+    {
+      baseTokenAddress: '0xd8763cba276a3738e6de85b4b3bf5fded6d6ca73',
+      debtTokenAddress: '0xfb6c830c13d8322b31b282ef1fe85cbb669d9ae8',
+      imageUrl:
+        'https://raw.githubusercontent.com/valora-inc/address-metadata/main/assets/tokens/mcEUR.png',
```
dievazqu: I couldn't find a better image for these tokens, I hope this is ok for this PR.
But I think we should use different images for these, maybe adding some red border (to indicate a debt) to those images? Should I ask Reed to help us? Or maybe someone knows something that we can already use here.

jeanregisser: For now you can use the Moola image for all of them. Like we do for Ubeswap. This may evolve in the future but is what Reed wanted.

src/apps/moola/plugin.e2e.ts
```diff
@@ -0,0 +1,12 @@
+import plugin from './plugin'
+
+describe('getPositionDefinitions', () => {
+  it('should get the address definitions successfully', async () => {
+    const positions = await plugin.getPositionDefinitions(
+      'celo',
+      '0x2f43b69f4ea32be168f19b998aaf5c3d87dbbdd6',
```
jeanregisser, line 7: I've updated our test account (used in other plugins) to have some debt too.
I'll share it with you too. Makes it easy for us to test all plugins.
```suggestion
      '0x2b8441ef13333ffa955c9ea5ab5b3692da95260d',
```

src/apps/moola/debtTokens.ts
```diff
@@ -0,0 +1,76 @@
...
+  title: string
+  description: string
+}
+
+export const MOOLA_DEBT_TOKENS: DebtTokenDefinition[] = [
+  {
+    baseTokenAddress: '0xd8763cba276a3738e6de85b4b3bf5fded6d6ca73',
+    debtTokenAddress: '0xfb6c830c13d8322b31b282ef1fe85cbb669d9ae8',
+    imageUrl:
+      'https://raw.githubusercontent.com/valora-inc/address-metadata/main/assets/tokens/mcEUR.png',
```
jeanregisser, line 16: We can use the Moola logo for all of them for now.
```suggestion
    imageUrl:
      'https://raw.githubusercontent.com/valora-inc/dapp-list/main/assets/moola.png',
```

jeanregisser (approved):
Amazing!! Thank you sir 🚀 🚀 🚀 
Love the simplicity of the implementation.

You're officially our very first hook contributor 🎉

src/apps/moola/plugin.e2e.ts
dievazqu: Thanks!

PR #120, opened 2023-07-05
feat!: add hooks api

jeanregisser:
This moves https://github.com/valora-inc/hooks-api to here.
It allows to easily run the api locally for previewing in Valora.
It will also be simpler to apply breaking changes.

The NPM publish step becomes unnecessary.

Fixes RET-750


src/api/staging.yaml
```diff
@@ -0,0 +1,2 @@
+# This is interestingly not set for us in the Cloud Function environment
```
silasbw, line 1: I've been calling this "testnet" (vs "staging").  Worth being consistent?  "testnet" seems like the convention elsewhere (e.g., blockchain platform providers).

.env.example
```diff
@@ -0,0 +1 @@
+GOOGLE_CLOUD_PROJECT=celo-mobile-alfajores
```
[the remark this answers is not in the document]
silasbw: Is for logging?  It's probably confusing to imply running locally requires a GOOGLE_CLOUD_PROJECT.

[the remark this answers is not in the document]
jeanregisser: Good point I'll remove it.

src/api/staging.yaml
jeanregisser: Ah good point, but testnet feels like too restricting, as if getting mainnet data wouldn't be possible with it.

What I have in mind is this:
- there would be 2 main builds for the wallet: production and staging
- both would be able to interact with Celo mainnet or other networks
- the main difference is they would consume data from the production or staging environments. The network would only be a parameter.

Alchemy doesn't use the testnet wording, they reference [networks](https://github.com/alchemyplatform/alchemy-sdk-js/blob/main/src/types/types.ts#L76):
https://docs.alchemy.com/reference/alchemy-gettokenbalances
QuickNode does the [same](https://www.quicknode.com/docs/celo#celo-network-support).

Let's chat about this.

silasbw: I probably don't know what this file is for, but it specifically references alfajores (instead of mainnet), so I don't understand "both would be able to interact with Celo mainnet or other networks".

More generally, "testnet" (or "mainnet") is a simpler and more scalable way to say alfajores-goerli-mumbai.  Are we going to change the environment names every time we add a new network?

We already have a convention around this and re-defining/changing the convention seems out of scope for this PR, but happy to keep chatting if it's important to redefine and update elsewhere now.

.env.example
```diff
@@ -0,0 +1,2 @@
+# This is only needed for real deployments to GCP
```
silasbw, line 1: Can we eliminate this option entirely for local development?  It's an odd valora production environment detail to leak into a builders dev workflow.

src/api/staging.yaml
silasbw: ...or, what's the case where you'd configure both alfajores and mainnet in the same environment?

.env.example
jeanregisser: yes removed.

src/api/staging.yaml
jeanregisser: This isn't something new introduced in this PR.
I first discussed it here: https://github.com/valora-inc/hooks-api/pull/3
The Google project name currently references the network name, but the goal would be to stop doing that.

silasbw (approved):

PR #121, opened 2023-07-05
feat: script with QR code to enable preview mode in Valora

jeanregisser:
This is the script to start the local server and show the QR code to enable preview mode in Valora

Screenshot 2023-07-05 at 19 17 10

Fixes RET-750


scripts/start.ts
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  const codes = colors[color]
+  if (codes) {
+    return '\x1B[' + codes[0] + 'm' + str + '\x1B[' + codes[1] + 'm'
+  }
+  return str
+}
+
+void (async () => {
```
silasbw, line 60: the `main().catch(err => { whatever }).then()` pattern we usually follow seems cleaner because it doesn't leave unresolved promises.  Is there a reason to not use that?

scripts/start.ts
```diff
@@ -0,0 +1,134 @@
...
+        if (!serverReadyOnce) {
+          printServerInfo()
+        } else {
+          console.log('› Server restarted and ready')
+        }
+        serverReadyOnce = true
+      }
+    })
+
+    child.stderr!.on('data', function (data: Buffer) {
```
silasbw, line 121: Can you set the child process to inherit std{in,out,err} from the parent?

scripts/start.ts
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  function startServer(command: string) {
+    const child = $.exec(command, { async: true, silent: true })
+    let serverReadyOnce = false
+    // This string is printed by the functions-framework when the server is ready
+    const serverReadyString = `URL: http://localhost:${argv.port}/`
+
+    // Wait for the server to be ready before printing the server info
+    child.stdout!.on('data', function (data: Buffer) {
```
silasbw, line 97: wouldn't, like, a TCP open or HTTP HEAD request or something be easier for checking if the server is ready w/o handling stdout events and buffer, etc?

scripts/start.ts
```diff
@@ -0,0 +1,134 @@
...
+  green: [32, 39],
+  magenta: [35, 39],
+  red: [31, 39],
+  yellow: [33, 39],
+}
+
+/**
+ * Add colors to the specified string, for colorized output in terminals
+ */
+function stylize(str: string, color: keyof typeof colors) {
```
silasbw, line 48: this is making some assumptions about terminals.  consider using https://www.npmjs.com/package/supports-color.

scripts/start.ts
jeanregisser: I wanted to modify the child process output to remove the reference to `http://localhost:{port}`, that's why I didn't let it output anything.

scripts/start.ts
jeanregisser: Not really because I wanted to control exactly when I'd insert the custom server info and replace some part of the default one. Waiting for readiness wasn't the main thing.

scripts/start.ts
jeanregisser: Good point 👍

scripts/start.ts
jeanregisser: Ah, I don't actually need top level async/await so I removed it.

scripts/start.ts
jeanregisser: I've switched to https://github.com/chalk/chalk, much better.

silasbw (approved):
:100:

PR #132, opened 2023-07-14
feat: good ghosting position using smart contract

camolezi:

PR #139, opened 2023-07-17
feat: isolate failures when getting positions from hooks

jeanregisser:
i.e. a failure getting position definitions from a specific hook won't prevent other hooks from returning their positions

Follow up from https://github.com/valora-inc/hooks-api/pull/80
This effectively re-enables HaloFi, now that it's been fixed in https://github.com/valora-inc/hooks/pull/132
and will prevent future failures from affecting positions from other hooks.


src/runtime/getPositions.ts
```diff
@@ -305,9 +306,20 @@ export async function getPositions(
...
+          return definitions.map((definition) => addAppId(definition, appId))
+        },
+        (err) => {
+          // In case of error, log and return empty array
+          // so other positions can still be resolved
+          logger.warn(
+            { err },
+            `Failed to get position definitions for ${appId}`,
+          )
+          return []
```
jeanregisser, line 320: This is the interesting change.

src/runtime/getPositions.ts
```diff
@@ -305,9 +306,20 @@ export async function getPositions(
...
-        return definitions.map((definition) => addAppId(definition, appId))
-      }),
+      plugin.getPositionDefinitions(network, address).then(
+        (definitions) => {
+          return definitions.map((definition) => addAppId(definition, appId))
+        },
+        (err) => {
+          // In case of error, log and return empty array
+          // so other positions can still be resolved
+          logger.warn(
```
kathaypacific, line 316: should we make this an error so that we get alerted about it?

kathaypacific (approved):
🙏🏻

src/runtime/getPositions.ts
jeanregisser: Good point. I've updated it to the `error` severity. We won't automatically get alerted by it though, since the API call will succeed.
But we should probably add one specific alert for this.

PR #141, opened 2023-07-18
fix: Halofi hook should filter withdrawn games

camolezi:
Bug fix and optimization for the halofi hook:

- Getting token decimals from "resolvedTokens" instead of querying on chain data.
- Bug fix: Filter Halofi games where the user has withdrawn


jeanregisser (approved):
Thank you 🙏

PR #194, opened 2023-09-08
fix(ubeswap): missing image since it was removed from dapp-list

jeanregisser:
Following the delisting of Ubeswap in https://github.com/valora-inc/dapp-list/pull/510
The related positions image was broken.

**before/after**

### DIFF
--- a/src/apps/ubeswap/positions.ts
+++ b/src/apps/ubeswap/positions.ts
@@ -70,7 +70,7 @@ async function getPoolPositionDefinition(
         title: `${token0.symbol} / ${token1.symbol}`,
         description: 'Pool',
         imageUrl:
-          'https://raw.githubusercontent.com/valora-inc/dapp-list/main/assets/ubeswap.png',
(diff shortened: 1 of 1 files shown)


silasbw (approved):

src/apps/ubeswap/positions.ts
```diff
@@ -70,7 +70,7 @@ async function getPoolPositionDefinition(
         title: `${token0.symbol} / ${token1.symbol}`,
         description: 'Pool',
         imageUrl:
-          'https://raw.githubusercontent.com/valora-inc/dapp-list/main/assets/ubeswap.png',
+          'https://raw.githubusercontent.com/valora-inc/dapp-list/ab12ab234b4a6e01eff599c6bd0b7d5b44d6f39d/assets/ubeswap.png',
```
MuckT, line 73: Should hooks have it's own assets list or should we make a shared asset list?

PR #197, opened 2023-09-11
feat!: make `getAppTokenDefinition` optional

jeanregisser:
`getAppTokenDefinition` is only needed in 2 situations:
1. when the token definition returns intermediary tokens which are not base tokens
2. when a hook depends on another hook's tokens which are not base tokens

Though 2 is not yet fully supported as we haven't had the case yet, but it exists. For instance Revo tokens depending on Ubeswap LP tokens.

So far only Ubeswap and Curve really needed to implement it.
This change aims to make the common use case simpler to implement for hooks author.

The runtime will through an useful error instead when `getAppTokenDefinition` is needed but no implemented.

### DIFF
--- a/src/runtime/getPositions.ts
+++ b/src/runtime/getPositions.ts
@@ -370,6 +370,11 @@ export async function getPositions(
             // TODO: We'll probably need to allow hooks to specify the app id themselves
             const { sourceAppId } = tokenDefinition
             const hook = hooksByAppId[sourceAppId]
+            if (!hook.getAppTokenDefinition) {
+              throw new Error(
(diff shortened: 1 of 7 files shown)


src/runtime/getPositions.ts
```diff
@@ -370,6 +370,11 @@ export async function getPositions(
             // TODO: We'll probably need to allow hooks to specify the app id themselves
             const { sourceAppId } = tokenDefinition
             const hook = hooksByAppId[sourceAppId]
+            if (!hook.getAppTokenDefinition) {
+              throw new Error(
```
dievazqu, line 374: Hmm, thinking twice about this...
Throwing here will avoid resolving other tokens definitions, right?
So we could be missing some positions?

jeanregisser: Yes it will fail the whole resolution.

It may not be ideal, but I think it's ok for now, as I prefer that this doesn't go unnoticed.

PR #200, opened 2023-09-19
feat: Add Uniswap positions hook

gnardini:
Added Uniswap position hook. It doesn't fetch the claimable fees data. Could be added as a shortcut to claim them next.


jeanregisser (approved):
OMG love this!!! ❤️  🚀 🚀 🚀 

Your multicall contract makes the implementation so succint!! 💪

PR #227, opened 2023-10-12
feat: getShortcutDefinitions takes network and address

silasbw:
In some cases the ShortcutDefinitions might be a function of network and address. For example, the hedgey a shortcut ID might the tokenID (which is a function of network and address).


silasbw:
This isn't a working example, but consider it a proposal for updating the runtime.  I can go about making changes to get it working and figure out backwards compatibility if the high-level changes are OK.

silasbw:
closing for https://github.com/valora-inc/hooks/pull/271

PR #271, opened 2023-11-27
feat: getShortcuts (optionally) takes network and address

silasbw:
New wallet implementations can use the v2 endpoint. Old wallet implementations will continue to work. New shortcuts can require (network, address), but they won't show on old wallet implementations.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -60,7 +60,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: google-github-actions/auth@v1
+      - uses: google-github-actions/auth@v2
         with:
           project_id: celo-mobile-alfajores
           credentials_json: ${{ secrets.ALFAJORES_SERVICE_ACCOUNT_KEY }}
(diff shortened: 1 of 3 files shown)


silasbw:
example of new shortcut implementation here: https://github.com/valora-inc/hooks/pull/227/files#diff-4835a7ed6db6e3c11e4bc4acec639412877cce313fd9937d66a4b487550f3b82

src/api/index.ts
```diff
@@ -27,7 +27,7 @@ function createApp() {
     }),
   )
 
-  const balancesRequestSchema = z.object({
+  const getHooksRequestSchema = z.object({
```
jeanregisser, line 30: 👍

scripts/getShortcuts.e2e.ts
```diff
@@ -2,7 +2,9 @@ import * as $ from 'shelljs'
 
 describe('getShortcuts', () => {
   it('should get shortcuts successfully', () => {
-    const result = $.exec('yarn getShortcuts')
+    const result = $.exec(
+      'yarn getShortcuts --address 0x2b8441ef13333ffa955c9ea5ab5b3692da95260d',
+    )
     expect(result.code).toBe(0)
   })
```
jeanregisser, line 9: Could we have 2 tests? one with and one without address?

jeanregisser (approved):
Great! 👍

PR #277, opened 2023-12-09
refactor(getShortcutDefinitions): define to return a promise

silasbw:
The runtime already resolves the value .getShortcutDefinitions returns.

### DIFF
--- a/src/apps/gooddollar/shortcuts.ts
+++ b/src/apps/gooddollar/shortcuts.ts
@@ -10,7 +10,7 @@ const client = createPublicClient({
 
 const hook: ShortcutsHook = {
   getShortcutDefinitions(_network?: string, _address?: string) {
-    return [
+    return Promise.resolve([
       {
         id: 'claim-reward',
         name: 'Claim',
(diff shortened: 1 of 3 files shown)


silasbw:
There's this guy in the pipeline, fyi: https://github.com/valora-inc/hooks/pull/278

src/apps/gooddollar/shortcuts.ts
```diff
@@ -10,7 +10,7 @@ const client = createPublicClient({
 
 const hook: ShortcutsHook = {
   getShortcutDefinitions(_network?: string, _address?: string) {
-    return [
+    return Promise.resolve([
```
jeanregisser, line 13: This works, but why not turning adding the `async` keyword to the function instead?
Feels simpler.

jeanregisser (approved):

src/apps/gooddollar/shortcuts.ts
silasbw: Meh, it felt superfluous because we don't use the await keyword.  I can change it.

PR #278, opened 2023-12-09
feat: add hedgey shortcut for claiming

silasbw:


jeanregisser (commented):
Awesome! 

Would it be possible to add an e2e test for the shortcut? Just to check we can get the full definition.

jeanregisser (approved):
🙌

PR #302, opened 2024-01-09
chore: add an example position hook

silasbw:


src/apps/example/positions.ts
```diff
@@ -0,0 +1,85 @@
...
+      toDecimalNumber(parseEther('1'), 18),
+      toDecimalNumber(parseEther('1'), 18),
+    ]
+
+    return [
+      {
+        // Position Hooks type, see https://docs.valora.xyz/hooks/
+        type: 'contract-position-definition',
+        network,
+        // TODO: What's the meaning of this property?
```
jeanregisser, line 67: I've documented it as "the address of the contract for the position definition", 
Maybe not clear enough but it should be the contract you interact with to get into that position.

For instance, when locking celo, it's the address of the locked gold contract.

silasbw:
A real work of art :laughing: 

Screenshot_20240109-103629

src/apps/example/positions.ts
```diff
@@ -0,0 +1,87 @@
...
+          imageUrl: DEFAULT_IMG_URL,
+        },
+      },
+    ]
+  },
+
+  async getAppTokenDefinition(_) {
+    throw new Error('Not implemented')
+  },
+}
```
jeanregisser, line 85: We can remove that one, it's not needed in most cases.

jeanregisser (approved):
🚀

PR #323, opened 2024-01-30
chore: API config option to specify APP_IDS

silasbw:
We don't want to enable all APP_IDS by default on staging and production.

For example, we'd like to avoid:

  https://github.com/valora-inc/hooks/pull/304


jeanregisser (commented):
Oh thanks, but I was thinking we'd be better off having an ignore list instead?

This way we avoid having to think about declaring explicitely each new app.
Assuming the main use case is mostly for ignoring specific apps (in progress, or the example app).

jeanregisser (approved):

PR #349, opened 2024-03-11
refactor(multichain)!: support other chains, starting with curve hook and ethereum

cajubelt:
Changes made:
- updates the api to use Valora's standard network ID's for multichain requests
- updates curve hook to actually return some data for multichain requests
- seals off all other hooks with comments indicating whether they can be extended or are specific to Celo
- uses blockchain api endpoint to get token info (which returns multichain token metadata and prices, rather than just celo token metadata and prices)
- refactors to use Valora's standard for token ID's instead of addresses as token identifiers, to work for positions that include native assets lacking a contract address, like ETH

BREAKING CHANGE: 'network' becomes 'networkId' in a few places in the API response. Requests can still use legacy 'network' request parameter, with either 'celo' or 'celoAlfajores' recognized. The wallet is updated in [this PR](https://github.com/valora-inc/wallet/pull/5093) to adapt to the change, but old versions should continue to work without changes, since the 'network' request is not used for anything critical (see description on [wallet PR](https://github.com/valora-inc/wallet/pull/5093) for details)

test plan:
- [x] **Backwards compatibility test**: run manual e2e test with 'main' branch of wallet (which still uses 'network' in requests and expects 'network' in response data) to verify that positions are queried for and displayed correctly for legacy app versions, e.g. that the changes are backwards compatible enough not to break anything. use local hooks endpoint, local wallet build, and hardcoded wallet address in the request.
Simulator Screen Shot - iPhone 14 - 2024-03-20 at 16 48 20
image
- [x] hunt thru wallet code to make sure the legacy 'network' field on some response data is not used for anything important
- [x] **Multichain functionality test**: run manual e2e test with `cajubelt/multichain-positions` branch of wallet, with an account that has a curve position on ethereum. use local hooks endpoint, local wallet build, and hardcoded network id in the request.
simulator_screenshot_6D05642F-8186-429F-86B1-980122A8FBC4
- [x] update existing unit and e2e tests to add coverage for curve on ethereum. sent the address used in e2e tests a curve lp token. 

cc https://linear.app/valora/issue/RET-941/%5Bhooks%5D-multichain-support


scripts/start.e2e.ts
```diff
@@ -1,9 +1,18 @@
 import * as $ from 'shelljs'
 import terminate from 'terminate/promise'
+import yaml from 'js-yaml'
+import fs from 'fs'
 
 describe('start', () => {
   it('should start the preview server successfully', async () => {
-    const child = $.exec('yarn start', { async: true })
+    const envFileContent = yaml.load(
```
cajubelt, line 8: this change was necessary because I added the first config parameter that was actually requires to be set, and did not default to something in code. this has the nice side effect of adding some coverage for our production config file, since this test will fail if there is a required config that is not included there

src/api/index.test.ts
```diff
@@ -6,9 +6,17 @@ import { getTestServer } from '@google-cloud/functions-framework/testing'
...
+import { NetworkId } from './networkId'
+import { getConfig } from './config'
+jest.mock('./config')
+jest.mocked(getConfig).mockReturnValue({
+  POSITION_IDS: [],
+  GET_TOKENS_INFO_URL: 'https://valoraapp.com/mock-endpoint',
+  GOOGLE_CLOUD_PROJECT: 'dev-project',
+  SHORTCUT_IDS: [],
+})
+import './index' // NOTE: there are side effects of importing this module-- loading config params from the environment in particular. so mocking configs MUST be done before importing.
```
cajubelt, line 19: this effectful imports issue is not new to this PR, though I strongly recommend we fix it soon and avoid it in the future

src/apps/hedgey/shortcuts.ts
```diff
@@ -3,15 +3,16 @@ import { celo } from 'viem/chains'
...
+import { NetworkId } from '../../api/networkId'
 
 const client = createPublicClient({
   chain: celo,
   transport: http(),
 })
 
 const hook: ShortcutsHook = {
-  async getShortcutDefinitions(network?: string, address?: string) {
-    if (network !== 'celo' || !address) {
```
cajubelt, line 14: @jeanregisser should the other shortcuts also check if the network id is defined and equal to Celo if that's the only network the dapp is on? I tried adding the check everywhere, but found that it broke some e2e tests in `runtime/getShortcuts.e2e.ts` that explicitly pass in "undefined" for the network. The tests seem to indicate that missing network id's are supposed to be accepted in general. In fact, the network id is ignored for the most part (except for locked celo), meaning you could query for ethereum shortcuts and get a bunch of celo results. 

I tried out using the network id parameter to compare against arrays of supported networks for each dapp in commit `4a94b15a34817f36e68cf6aaa89691475724fbf0`, but ran up against the e2e test failure. 

Lmk what the intent here and how you think we should proceed

cajubelt: it looks like the 'network' parameter was required in the api, which makes me think maybe the e2e test enforcing the acceptance of undefined network parameters was not testing something that's important to preserve, but lmk if I'm mistaken about that

src/apps/ubeswap/positions.ts
```diff
@@ -57,23 +60,43 @@ async function getPoolPositionDefinition(
...
+        resolvedTokensByTokenId[
+          getTokenId({
+            networkId,
+            address: token1Address,
+            isNative: isNative({
+              networkId: networkId,
+              address: token1Address,
+            }),
+          })
+        ]
```
jeanregisser, line 91: nit: I wonder if we could find a way to keep this short.

Maybe `isNative` is a bit redundant? 
Or I guess we could have this function be called with a helper function that's more succint? `getResolvedToken(networkId, address)` which would use `resolvedTokensByTokenId` under the hood.

Feel free to ignore here, we can do that in a follow up PR.

src/runtime/client.ts
```diff
@@ -0,0 +1,30 @@
...
+  [NetworkId['arbitrum-sepolia']]: arbitrumSepolia,
+  [NetworkId['op-sepolia']]: optimismSepolia,
+}
+
+export function getClient(networkId: NetworkId) {
+  return createPublicClient({
+    chain: networkIdToViemChain[networkId],
+    transport: http(),
+  })
+}
```
jeanregisser, line 30: Does this erase the actual typed chain? Not sure it matters though.

src/runtime/getPositions.ts
```diff
@@ -153,33 +170,52 @@ function getDisplayProps(
...
+      tokensByTokenId[
+        getTokenId({
+          networkId: token.networkId,
+          address: token.address,
+          isNative: isNative({
+            networkId: token.networkId,
+            address: token.address,
+          }),
+        })
+      ]
```
jeanregisser, line 199: Seeing this code written multiple times, I think I'm now convinced we should address it in this PR and make it more succint to get a token from a network/address pair.

Maybe with a helper?

src/runtime/getTokenId.ts
```diff
@@ -0,0 +1,13 @@
...
+  networkId,
+  isNative,
+  address,
+}: {
+  networkId: NetworkId
+  isNative: boolean
+  address?: string
+}) {
+  return `${networkId}:${isNative ? 'native' : address?.toLowerCase()}`
+}
```
jeanregisser, line 13: Nothing to do now, but I think it's the 2nd time this week I notice we spread this helper to multiple repos.
Maybe we should have a small lib to avoid the duplication.

src/runtime/isNative.ts
```diff
@@ -0,0 +1,26 @@
...
+    return (
+      address.toLowerCase() === '0x471ece3750da237f93b8e339c536989b8978a438'
+    )
+  }
+  if (networkId === NetworkId['celo-alfajores']) {
+    return (
+      address.toLowerCase() === '0xf194afdf50b03e69bd7d057c1aa9e10c9954e4c9'
+    )
+  }
+  return address.toLowerCase() === ETHER_HEX_IDENTIFIER
```
jeanregisser, line 25: Next time we add a network, we'll need to remember to add another case here.
Can we have a compile time check that would fail if we forget?

jeanregisser (commented):
Awesome work!! 🚀 🚀 🚀

jeanregisser:
What do you think of updating the PR title to reflect that it's actually making the whole repo support multichain?

src/api/networkId.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export enum NetworkId {
+  'celo-mainnet' = 'celo-mainnet',
+  'celo-alfajores' = 'celo-alfajores',
+  'ethereum-mainnet' = 'ethereum-mainnet',
+  'ethereum-sepolia' = 'ethereum-sepolia',
+  'arbitrum-one' = 'arbitrum-one',
+  'arbitrum-sepolia' = 'arbitrum-sepolia',
+  'op-mainnet' = 'op-mainnet',
+  'op-sepolia' = 'op-sepolia',
```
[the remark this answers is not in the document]
cajubelt: I was on the fence about it but went with op-mainnet/sepolia because it aligns better with how Optimism's docs talk about their chains https://docs.optimism.io/chain/getting-started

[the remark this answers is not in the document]
cajubelt: I made this choice earlier when we first introduced optimism in other repos btw (not in this PR). Though it wouldn't be too late to change it unless you guys have released it since then

src/runtime/getPositions.ts
cajubelt: yeah it's ugly, and not safe from bugs this way since you could accidentally pass a different address or networkId in to the two functions (getTokenId and isNative) by mistake, for the same token. I updated how getTokenId is working to call isNative itself when we aren't sure either way if a token is native or not. lmk what you think.

src/runtime/getTokenId.ts
cajubelt: yeah makes me nervous every time lol. It would be easy to see some drift, particularly in whether the address is lowercased or not

src/runtime/isNative.ts
cajubelt: I was about to write "but other chains won't work differently" and then I was like nah he's right lol.

src/api/networkId.ts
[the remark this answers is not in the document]
cajubelt: done

src/apps/ubeswap/positions.ts
cajubelt: I tried out making it more succinct by calling isNative from getTokenId

src/apps/hedgey/shortcuts.ts
cajubelt: ok, since no shortcuts are multichain right now I'm ok with leaving it for when that functionality is added then

src/runtime/client.ts
cajubelt: yeah I'm not sure either

cajubelt: I tinkered around with it and didn't come up with any useful improvements

cajubelt: the best I thought of was setting the return type to `PublicClient<ReturnType<typeof http>, Chain>` explicitly, but that still lacks specificity on the chain type, and idk if it would help us

cajubelt: my sense is that the more generic return type we'll get from getClient should throw type errors if we ever try to access some network-specific functionality that isn't present for all chains. so we'll have a chance to deal with it if it comes up. but lmk if you're thinking of something else

jest.config.js
```diff
@@ -1,7 +1,10 @@
 /** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
+require('ts-node/register') // allows us to use typescript for setup script
+require('./scripts/loadProductionEnvVars.ts')
 
 module.exports = {
   projects: ['<rootDir>/jest.unit.config.js', '<rootDir>/jest.e2e.config.js'],
+  setupFiles: ['<rootDir>/scripts/loadProductionEnvVars.ts'],
```
jeanregisser, line 7: nit: Is it possible to do this only from `jest.e2e.config.js`?
It's only needed for e2e tests, right?

src/api/config.ts
```diff
@@ -1,11 +1,44 @@
...
+  POSITION_IDS: string[]
+  SHORTCUT_IDS: string[]
+}
 
+export function networkIdToRpcUrlTransform(val: string | undefined) {
+  if (!val) {
+    return {}
+  }
+  // expected format: network id/rpc url pairs joined by |, and pairs separated by spaces
+  // example: 'ethereum-mainnet|https://my-endpoint-name.quiknode.pro/my-api-key celo-mainnet|https://forno.celo.org'
```
jeanregisser, line 18: nit: wondering if it wouldn't be easier to have a multiline KEY=VALUE config instead? At least for editing the secret in secrets manager.

src/runtime/getPositions.ts
jeanregisser: Thanks, this looks good now 👍

jeanregisser (approved):
🙌 

Already approved with some very small nits.

src/api/config.ts
cajubelt: something this?
```
ethereum-mainnet=https://my-endpoint-name.quiknode.pro/my-api-key
celo-mainnet=https://forno.celo.org
```

I didn't use equals as a separator because it's commonly used in URLs, so splitting on `'='` would break. Lmk if you were thinking of something else though

cajubelt: we could certainly use newlines instead of spaces, but I used spaces because I was doing some testing in the CLI and spaces are more convenient there

cajubelt: I'm going to merge for now, but lmk if you were thinking of something else and I'm happy to circle back

PR #350, opened 2024-03-13
chore(curve): update to use non-deprecated curve pools endpoint

cajubelt:
precursor to https://linear.app/valora/issue/RET-941/[hooks]-multichain-support which I am doing on a contract basis

I noticed this when updating hooks to be multichain, since when I went to add values to the `CURVE_POOL_URLS` object I noticed that the urls were for a deprecated endpoint

test coverage:
- new unit tests in curve/positions.test.ts
- existing e2e test in curve/positions.e2e.ts

previous endpoint was deprecated, see [here](https://api.curve.fi/v1/documentation/#/Deprecated/get_getFactoryV2Pools__blockchainId_)

api docs for new endpoint [here](https://api.curve.fi/v1/documentation/#/Pools/get_getPools__blockchainId___registryId_)


src/apps/curve/positions.test.ts
jeanregisser, line 1: I wanted to experiment favoring e2e tests in this repo as much as possible. And using unit tests only for cases that can't be easily covered by an e2e test.

Could we drop this one?

src/apps/curve/positions.test.ts
```diff
@@ -0,0 +1,244 @@
...
+jest.mock('got', () => ({
+  get: jest.fn().mockImplementation((url: string) => {
+    if (url === 'https://api.curve.fi/v1/getPools/celo/factory') {
+      return {
+        json: jest.fn().mockResolvedValue(mockResponse),
+      }
+    }
+    throw new Error('Invalid curve url')
+  }),
+}))
```
jeanregisser, line 12: nit: this repo uses https://mswjs.io/ to mock API requests.

This way mocks are not dependant on a specific network request library.

Mock server is in https://github.com/valora-inc/hooks/blob/main/test/server.ts

jeanregisser (commented):
Great, thanks!

src/apps/curve/positions.test.ts
cajubelt: thanks, sounds good. i refactored to use msw for mocking

src/apps/curve/positions.test.ts
cajubelt: given that the unit test is written, this would mean adding a commit specifically to drop a test. In my view, unit tests should outweigh e2e tests because they're easy to write, easy to debug, and helpful to isolate problems when bugs are introduced. If this one becomes annoying to maintain or causes some other problem, you are welcome to drop it-- but I don't think it's an improvement to drop it right now.

silasbw: what's the experiment you have in mind?

cajubelt: if you have an idea for how this can be helpfully morphed into an e2e test though, I'm open to it. I wrote it as a unit test because I guessed that the data returned by Curve's endpoint could change, so our e2e tests might need to be too inspecific. And even if that isn't the case, it takes time and money to prep a wallet address to have certain positions on mainnet, and can be annoying to update those accounts to have other positions later on

silasbw (approved):

PR #362, opened 2024-04-04
refactor(multichain)!: accept multiple network ids for get positions and get shortcuts

cajubelt:
BREAKING CHANGE: instead of singular "networkId", takes array "networkIds" as query parameter for endpoints to get positions and shortcuts. The "networkId" query param was added recently (by me) and not used by any wallet versions, so this should be ok.

In my last hooks PR, I assumed we would want to keep the api more or less intact, just substituting "networkId" for "network" since it expected different values. However, when I went to implement the wallet changes, I noticed that we'd need to make a separate request to the hooks endpoints for every network we want to support, which doesn't scale well. Taking multiple network ID's as a query parameter seems like the better approach. 

tested manually with local wallet build, verified that Celo and Eth positions can be displayed together. 

Simulator Screen Shot - iPhone 13 - 2024-04-04 at 10 59 18

### DIFF
--- a/src/api/index.test.ts
+++ b/src/api/index.test.ts
@@ -22,7 +22,7 @@ import './index' // NOTE: there are side effects of importing this module-- load
 jest.mock('../runtime/getPositions')
 jest.mock('../runtime/getShortcuts')
 
-const TEST_POSITIONS: Position[] = [
+const TEST_POSITIONS_CELO: Position[] = [
   {
     type: 'app-token',
     appId: 'ubeswap',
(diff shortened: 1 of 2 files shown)


src/api/index.test.ts
```diff
@@ -100,24 +157,39 @@ jest.mocked(getShortcuts).mockResolvedValue(TEST_SHORTCUTS)
...
+      })
   })
 
   it('returns 400 when address is missing', async () => {
     const server = getTestServer('hooks-api')
     const response = await request(server)
       .get('/getPositions')
       .query({
-        networkId: 'celo-mainnet',
+        network: 'celo', // note: this old schema should still be supported. only the address should be reported as an error.
```
cajubelt, line 192:
```suggestion
        network: 'celo', // note: this old schema should still be supported. only the missing address should be reported as an error.
```

src/api/index.ts
```diff
@@ -54,16 +69,19 @@ function createApp() {
...
-          : parsedRequest.query.networkId
-      const positions = await getPositions(
-        networkId,
-        address,
-        config.POSITION_IDS,
-        config.GET_TOKENS_INFO_URL,
-      )
+      const networkIds = getNetworkIds(parsedRequest.query)
+      const positions = (
+        await Promise.all(
```
silasbw, line 74: Will this change the wallet UX (compared to making one request per network)?  What about the example where (for whatever reason) `getPositions` is broken for eth?  Would the one-request-per-network approach render positions for the other chains and just exclude eth ones (and the change here would cause the wallet to show no positions).

silasbw (approved):

src/api/index.ts
silasbw: ...if we think the one-request-per-network UX is better, we could consider tweaking this to PR to return whatever positions are working

cajubelt: Yes, if the hooks service fails to get positions for one network, the entire request would fail with my PR as written. As you've noted, if we did one request per network, we could handle failures differently on the client side, for instance by displaying just the positions which are successfully returned. I think that's the main downside with this approach. However, I thought the benefit of having only one network request for an arbitrary number of networks was worth it. My guess is that random network failures or latency would increase with the number of requests, so I'm suggesting 1-request-for-n-networks approach instead of 1-request-per-network

cajubelt: And yeah, errors for just one network could be swallowed on the backend too if that's the experience we want to support on the client

jeanregisser: Thanks, yeah this raises an interesting question on what we want to do moving forward.
For blockchain-api we did 1 request per network on the client side. I think I was initially more in favor of a single request for all networks, but considering failures and displaying data to the user more quickly, I'm leaning more toward letting the client do N requests.
Especially here in this project where querying all positions could take longer time, as we add more apps on Ethereum.
Maybe something to discuss at tomorrow's eng meeting?

jeanregisser (approved):

PR #386, opened 2024-05-29
fix(curve): support all curve pools

jeanregisser:
I deposited funds into [TriCRV-ARBITRUM](https://curve.fi/#/arbitrum/pools/factory-tricrypto-14/deposit) on curve and noticed the pool wasn't showing up in Valora.

Here's the fix.

Note: it doesn't show the $ price because we haven't yet declared those tokens as base tokens and so they don't have prices. Would be nice to still display the token amount though.


src/apps/curve/positions.ts
```diff
@@ -44,12 +47,12 @@ export async function getAllCurvePools(
...
   }
   const { data } = await got
-    .get(`https://api.curve.fi/v1/getPools/${blockchainId}/factory`)
+    .get(`https://api.curve.fi/v1/getPools/all/${blockchainId}`)
     .json<CurveApiResponse>()
 
   return data.poolData.map((poolInfo) => ({
-    address: poolInfo.address,
-    size: poolInfo.implementation === 'plain3basic' ? 3 : 2,
+    address: poolInfo.lpTokenAddress,
```
jeanregisser, line 54: Some pools have a different lp token address, and here we really want lp tokens.

satish-ravi (approved):

PR #401, opened 2024-06-13
feat: add fallbackPriceUsd as escape hatch

jeanregisser:
Related to https://github.com/valora-inc/hooks/pull/396 and discussed with @gnardini ([thread](https://valora-app.slack.com/archives/C04NFRMSZQX/p1718275117206789))

This should make it easier to implement Beefy which relies on a wide variety of underlying tokens for which we don't yet have apps that could decompose them.

### DIFF
--- a/src/runtime/getPositions.test.ts
+++ b/src/runtime/getPositions.test.ts
@@ -185,7 +185,7 @@ describe(getPositions, () => {
                 fallbackPriceUsd: toSerializedDecimalNumber(0.5),
               },
             ],
-            // Meaning that 1 share of the vault is worth 2 underlying token
+            // Meaning that 1 share of the vault is worth 2 underlying tokens
             pricePerShare: [toDecimalNumber(2n, 0)],
             displayProps: {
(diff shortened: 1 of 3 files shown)


src/runtime/getPositions.test.ts
```diff
@@ -151,4 +153,90 @@ describe(getPositions, () => {
...
+              {
+                // Underlying Curve LP token
+                address: '0x4456d13Fc6736e8e8330394c0C622103E06ea419',
+                networkId,
+                // Fallback until we can properly decompose the token into base tokens
+                fallbackPriceUsd: '0.5' as SerializedDecimalNumber,
+              },
+            ],
+            // Meaning that 1 share of the vault is worth 2 underlying token
+            pricePerShare: [toDecimalNumber(2n, 0)],
```
jeanregisser, line 189: @gnardini this should interest you for implementing beefy vaults as `app-token-definition`

gnardini (approved):
Looks great! I'll update the Beefy PR so that it uses this soon :)

PR #403, opened 2024-06-14
feat: add MENTO airdrop

jeanregisser:
Addresses eligible to the MENTO airdrop will see the following:

 

Note: MENTO doesn't have a $ price yet, that's why the wallet shows "-". Ideally it would show the number of MENTO units. We need to fix that on the wallet side.

Also it would be great if we supported tapping on positions and opening the website directly.


src/apps/mento/positions.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  const position: ContractPositionDefinition = {
+    type: 'contract-position-definition',
+    networkId,
+    address: AIRDROP_ADDRESS,
+    tokens: [{ address: tokenAddress, networkId }],
+    displayProps: {
+      title: 'MENTO Airdrop',
+      description: 'Claim on https://airdrop.mento.org before August 9th, 2024',
```
MuckT, line 71: Do we just remove this code on August 9th or is it fine to have an old hook in the repo?

src/apps/mento/positions.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  return position
+}
+
+const hook: PositionsHook = {
+  getInfo() {
+    return {
+      id: 'mento',
+      name: 'Mento',
+      description: '',
```
MuckT, line 96: Do we need a description here?

src/apps/mento/positions.ts
silasbw: Can we stop showing the position on, say, aug 10?  e.g., "`if (date >= aug 10) return undefined`"

silasbw (approved):

PR #412, opened 2024-06-19
feat: add Stake DAO on arbitrum-one

jeanregisser:
This allows seeing staked Curve LP tokens from Stake DAO.

 

Fixes RET-1086


silasbw (approved):

src/apps/stake-dao/positions.ts
```diff
@@ -0,0 +1,172 @@
...
+        abi: stakeDaoVaultAbi,
+        functionName: 'sdGauge',
+      },
+    ]),
+    allowFailure: false,
+  })
+
+  const vaults = vaultAddresses.map((address, i) => ({
+    address,
+    lpTokenAddress: (vaultsData[2 * i] as Address).toLowerCase() as Address,
```
bakoushin, line 71: supernit: it looks like number `2` is related to the number of calls we supply to the multicall above.
and we have to remember to keep it in sync if we alter the multicall.
perhaps we can provide the length of those distinct calls instead of the literal number?

PR #423, opened 2024-06-25
Add Beefy CLM Vaults support

gnardini:
Beefy recently created CLM (Cowcentraded Liquidity Manager) Vaults which work a bit differently so they aren't included in the main `/vaults` endpoint. They can be fetched from a similar endpoint which is not documented (they suggested I use it on Discord) called `/harvestable-vaults` which includes all the "normal" vaults plus the CLM ones.

They need to be handled a bit differently, since the token is an NFT, not an ERC20, so we use a `ContractPositionDefinition` instead.


src/apps/beefy/abis/BeefyClmVaults.sol
```diff
@@ -0,0 +1,44 @@
...
+
+contract BeefyClmVaults {
+    struct VaultInfo {
+        address token0;
+        address token1;
+        uint256 amount0;
+        uint256 amount1;
+    }
+
+    function getUserVaults(
```
gnardini, line 22: I wrote this custom multicall contract. Technically, this could have been a normal multicall, the reason I did it this way is that I find it easier to read and work with on-chain data this way rather than having to manage the indexes for each call and then join it together again (ie for. each vault we have 4 calls, so we need to make sure to read the results in the right way).

The negative side of this approach is that we need to compile the contract to get the bytecode. I compiled it by writing the contract in another repo I have and then just copy-pasting the .sol file here, but maybe that's not the best approach. I can do it the other way if you think this complicates things too much.

jeanregisser: This seems ok, good to experiment!
Later we could automatically compile and update the bytecode to make it easier.

src/apps/beefy/positions.ts
```diff
@@ -88,38 +184,34 @@ const hook: PositionsHook = {
...
     )
+    const info =
+      clmVaults.length === 0
+        ? []
+        : ((await client.readContract({
+            code: beefyClmVaultsMulticallBytecode,
+            abi: beefyClmVaultsMulticallAbi,
+            functionName: 'getUserVaults',
+            args: [address, clmVaults.map((vault) => vault.earnedTokenAddress)],
+          })) as ClmVaultBalanceInfo[])
```
jeanregisser, line 198: nit: is it possible to let TS infer the return type from the ABI?

jeanregisser (approved):
Awesome! 🚀

src/apps/beefy/abis/BeefyClmVaults.sol
jeanregisser: Doesn't matter now, and please don't change, but if I remember correctly, one small advantage of the standard multicall with viem is that viem splits it into multiple RPC calls if the `data` part becomes too big.

https://viem.sh/docs/clients/public#batchmulticallbatchsize-optional

gnardini: That looks like a nice feature for multicalls :)

src/apps/beefy/positions.ts
gnardini: I spent like 20 minutes the other day trying to do it and I couldn't figure it out, and today I realized I was missing the `as const` in the abi in under a minute 😂

PR #424, opened 2024-06-25
feat(ubeswap): Add tracking for Ubeswap V3 pools

gnardini:
Using the same logic we use for Uniswap V3 to track Ubeswap V3 pools.

### DIFF
--- a/src/apps/ubeswap/positions.e2e.ts
+++ b/src/apps/ubeswap/positions.e2e.ts
@@ -7,8 +7,18 @@ describe('getPositionDefinitions', () => {
       NetworkId['celo-mainnet'],
       '0x2b8441ef13333ffa955c9ea5ab5b3692da95260d',
     )
-    // Simple check to make sure we got some definitions
-    expect(positions.length).toBeGreaterThan(0)
+    // Simple checks to make sure we got some definitions
+
+    // ube v2 pool definitions
+    expect(
+      positions.filter((p) => p.type === 'app-token-definition').length,
+    ).toBeGreaterThan(0)
+
+    // ube v2 farm definitions and ube v3 pools are both contract definitions. We could
+    // call displayProps to distinguish v2 farms from v3 pools for better coverage.
+    expect(
+      positions.filter((p) => p.type === 'contract-position-definition').length,
+    ).toBeGreaterThan(0)
   })
 
   it('should get no definitions for an address with no blockchain interaction', async () => {
--- a/src/apps/ubeswap/positions.ts
+++ b/src/apps/ubeswap/positions.ts
@@ -17,6 +17,7 @@ import { stakingRewardsAbi } from './abis/staking-rewards'
 import farms from './data/farms.json'
 import { NetworkId } from '../../types/networkId'
 import { getTokenId } from '../../runtime/getTokenId'
+import { getUniswapV3PositionDefinitions } from '../uniswap/positions'
 
 const client = createPublicClient({
   chain: celo,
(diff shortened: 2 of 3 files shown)


src/apps/ubeswap/positions.ts
```diff
@@ -17,6 +17,7 @@ import { stakingRewardsAbi } from './abis/staking-rewards'
 import farms from './data/farms.json'
 import { NetworkId } from '../../types/networkId'
 import { getTokenId } from '../../runtime/getTokenId'
+import { getUniswapV3PositionDefinitions } from '../uniswap/positions'
```
jeanregisser, line 20: Initially I thought we should avoid inter-app imports like this, to make it easier to split this repo (if we ever get to this and we have so many apps it becomes necessary, or for a more permissionless version of hooks).
but I think here it's definitely the right thing to do.

src/apps/uniswap/positions.ts
```diff
@@ -53,6 +54,81 @@ const UNI_V3_ADDRESSES_BY_NETWORK_ID: {
...
+              getTokenId({
+                address: pool.token1,
+                networkId,
+              })
+            ].symbol
+          }`,
+          description: 'Pool',
+          imageUrl:
+            'https://raw.githubusercontent.com/valora-inc/dapp-list/ab12ab234b4a6e01eff599c6bd0b7d5b44d6f39d/assets/uniswap.png',
+        }),
```
jeanregisser, line 107: Looks like we should make this a parameter, so the Ubeswap V3 pools still use the Ubeswap logo, right?

src/apps/ubeswap/positions.e2e.ts
```diff
@@ -7,8 +7,18 @@ describe('getPositionDefinitions', () => {
...
-    expect(positions.length).toBeGreaterThan(0)
+    // Simple checks to make sure we got some definitions
+
+    // ube v2 pool definitions
+    expect(
+      positions.filter((p) => p.type === 'app-token-definition').length,
+    ).toBeGreaterThan(0)
+
+    // ube v2 farm definitions and ube v3 pools are both contract definitions. We could
+    // call displayProps to distinguish v2 farms from v3 pools for better coverage.
```
jeanregisser, line 18: Good idea! Do you mind doing it?

jeanregisser (approved):
Love this! 🚀

src/apps/ubeswap/positions.ts
gnardini: Another option would be to move shared stuff to a /shared directory or something like that, so apps can import from /shared but not other apps.

src/apps/uniswap/positions.ts
gnardini: Oops, my bad

src/apps/ubeswap/positions.e2e.ts
gnardini: Done, it looks a bit messy tbh I think, I think we could use the actual runtime for these types of tests in the future since we're reimplementing a lot of the logic that is already there (except mocking the actual data) to be able to call `displayProps`.

src/apps/ubeswap/positions.ts
```diff
@@ -138,14 +141,17 @@ async function getPoolPositionDefinitions(
 ): Promise<PositionDefinition[]> {
   // Get the pairs from Ubeswap via The Graph
   const { data } = await got
-    .post('https://api.thegraph.com/subgraphs/name/ubeswap/ubeswap', {
-      json: {
-        query: PAIRS_QUERY,
-        variables: {
-          address: address.toLowerCase(),
+    .post(
+      'https://gateway-arbitrum.network.thegraph.com/api/3f1b45f0fd92b4f414a3158b0381f482/subgraphs/id/JWDRLCwj4H945xEkbB6eocBSZcYnibqcJPJ8h9davFi',
```
gnardini, line 145: I found on the network tab of https://info.ubeswap.org/ they were using this graph now. Not sure how long it's going to last though, Ubeswap seems to be kind of inactive, even with their V3 launch.

src/apps/ubeswap/positions.e2e.ts
jeanregisser: Oh right yes would be better with with the runtime.

src/apps/ubeswap/positions.ts
jeanregisser: Thanks, this is good for now! Weird it says `gateway-arbitrum` though 👀

PR #429, opened 2024-07-01
feat(shortcuts): author defined trigger inputs

jeanregisser:
Up until now shortcuts were limited to the following input arguments:
- `networkId`
- `address`
- `positionAddress`

This PR allows arbitrary inputs when triggering shortcuts.
`networkId` and `address` are still always required, but any other argument is now up to the shortcut author.
`positionAddress` is now also entirely author defined.

This allows triggering shortcuts with a token and token amount for instance for the "earn" feature of Valora.
And allows for much richer interactions with apps.

Fixes RET-1086

### DIFF
--- a/src/apps/gooddollar/shortcuts.ts
+++ b/src/apps/gooddollar/shortcuts.ts
@@ -1,8 +1,9 @@
 import { Address, createPublicClient, http, encodeFunctionData } from 'viem'
 import { celo } from 'viem/chains'
-import { ShortcutsHook } from '../../types/shortcuts'
+import { createShortcut, ShortcutsHook } from '../../types/shortcuts'
 import { ubiSchemeAbi } from './abis/ubi-scheme'
 import { NetworkId } from '../../types/networkId'
+import { ZodAddressLowerCased } from '../../types/address'
 
 const client = createPublicClient({
   chain: celo,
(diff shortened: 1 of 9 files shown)


src/types/shortcuts.ts
```diff
@@ -1,22 +1,25 @@
...
   category?: 'claim' // We'll add more categories later
+  triggerInputShape: TriggerInputShape // Zod object shape of the input for the trigger
   onTrigger: (
-    networkId: NetworkId,
-    address: string,
-    positionAddress: string,
+    args: {
+      networkId: NetworkId
+      address: string
+    } & z.infer<ZodObject<TriggerInputShape>>,
```
jeanregisser, line 22: This is the main interesting change.

The additional trigger arguments are fully checked at runtime + compile time thanks to Zod.

src/apps/gooddollar/shortcuts.ts
```diff
@@ -16,16 +17,19 @@ const hook: ShortcutsHook = {
...
         id: 'claim-reward',
         name: 'Claim',
         description: 'Claim daily UBI rewards',
         networkIds: [NetworkId['celo-mainnet']],
         category: 'claim',
-        async onTrigger(networkId, address, positionAddress) {
+        triggerInputShape: {
+          positionAddress: ZodAddressLowerCased,
+        },
+        async onTrigger({ networkId, address, positionAddress }) {
```
jeanregisser, line 29: And an example with GoodDollar.

silasbw (approved):
ffffffooooooohhhh :fire:

PR #433, opened 2024-07-05
feat(compound): Add position tracking for Compound

gnardini:
Added position tracking for Compound markets.

In Compound, there is one contract for each borrowable token. For each contract, you can either supply the "base token" or supply one of the collateral tokens and then borrow the base token. This means each token's market is isolated from the rest, unlike Aave where you can use the same collateral to borrow several different tokens.

In any case, this particularity and the fact that Compound doesn't have "app tokens" made it difficult to fit into the existing API. If I create a single 'contract position' that includes the collateral + the debt then the app would show just the total value of |collateral| - |debt| but not the details of how many collateral tokens I have or how much debt I have. All of this info lives in the same contract address.

So what I did instead is to create many "contract positions" for each market, one for the base token supply amount, one for the debt and one for each collateral token, so users can see how much of each token they have. To do this, I had to put a non-address value on the `address` field:  
```
address: `${marketAddress.toLowerCase()}-${tokenAddress.toLowerCase()}-${type}`,
```
If I put only the market address then the positions collided and only one showed up in the app.

IMG_B1A7C726A64D-1
PS: I swapped "Borrow" for "Debt" after taking the screenshot


src/apps/compound/positions.e2e.ts
```diff
@@ -0,0 +1,41 @@
...
+    const collateralPosition = positions.find((p) =>
+      p.displayProps.title.endsWith(' Collateral'),
+    )
+    expect(collateralPosition?.tokens.length).toBeGreaterThan(0)
+
+    // TODO: Add debt position (requires access to the test account's private keys).
+    // const debtPosition = positions.find((p) =>
+    //   p.displayProps.title.endsWith(' Debt'),
+    // )
+    // expect(debtPosition?.tokens.length).toBeGreaterThan(0)
```
jeanregisser, line 31: I'll give you access!

src/apps/compound/positions.ts
```diff
@@ -0,0 +1,173 @@
...
+        const positions: ContractPositionDefinition[] = []
+
+        for (const { token, balance, market, type } of balances) {
+          if (balance === 0n) {
+            continue
+          }
+          positions.push({
+            type: 'contract-position-definition',
+            networkId,
+            address: `${market.toLowerCase()}-${token.toLowerCase()}-${type}`,
```
jeanregisser, line 122: Ah yes I really need to introduce a proper position identifier so we don't need this workaround.
This has been a problem for uniswap v3 and other apps too.
Related to shortcuts and the new earn feature in Valora, this is also on my TODO list for this week.

I thought Valora would still display individual positions, even with the same address, but maybe we broke that recently.

src/apps/compound/positions.ts
```diff
@@ -0,0 +1,173 @@
...
+          market: Address
+          balance: bigint
+          type: string
+        }[] = [
+          {
+            token: baseToken,
+            market: markets[index].address,
+            balance: baseTokenBalance,
+            type: 'Supply',
+          },
```
jeanregisser, line 95: Maybe this one could/should be modeled using `app-token-definition`? 
This way we would have a price for `cToken`?
i.e. the supplied tokens would show up with both a token amount and $ value in the positions.

src/apps/compound/assets/compound.png
jeanregisser, line 1: Thanks, I guess we should still include Compound in the dapp list too.
I'll add it.

src/apps/ubeswap/positions.e2e.ts
jeanregisser, line 1: oh this is cool! But could you split that into a separate PR?

jeanregisser (approved):
Awesome!! Great to this see! 🚀

src/apps/compound/positions.ts
gnardini: Ohh, right, I hadn't noticed token amounts don't show up.

I don't think we can use a token definition because the contract is not an ERC20. cTokens were used on Compound v2, but on the current live version (v3) there are no cTokens. The contract doesn't have `symbol` or `name` functions (it does seem to have the others though 🤔).

Maybe we can make it so if a contract position has only one token we can show the underlying token amount too? Or maybe create a different kind of position for one-token positions, not sure.

src/apps/ubeswap/positions.e2e.ts
gnardini: Opened this PR for it: https://github.com/valora-inc/hooks/pull/441

src/apps/compound/positions.e2e.ts
gnardini: Added!

src/apps/compound/positions.ts
jeanregisser: Oh ok, my knowledge cutoff was at Compound v2 😄 

Yes good idea, we should update Valora so it displays the token amount when there's only one. I wanted this for other contract positions.

PR #441, opened 2024-07-10
fix(tests): Improve Ubeswap e2e tests

gnardini:
I mocked a bunch of stuff in a previous PR instead of using the runtime for Ubeswap e2e tests. Switching to use the runtime to make the test much more readable.


jeanregisser (approved):
Thanks, this is great 👍

PR #442, opened 2024-07-10
fix(beefy): Bugfix for Beefy CLM Vaults

gnardini:
There was a bug in Beefy positions. The positions for CLM vaults were throwing an error because they were not being categorized as CLM vaults. To check, we were using the `strategyTypeId` field instead of the `type` field.

I'm not sure how this error happened, if they changed the values for `strategyTypeId` in the API or I made a mistake somewhere because this was working before but now isn't anymore. In any case, this small fix solved the issue.

I also added a test to cover for this case.


jeanregisser (approved):
🚀

PR #444, opened 2024-07-12
feat: add positionId

jeanregisser:
Attach a unique identifier to positions.
- The format used is the following: `{networkId}:{positionAddress}(:{optionalExtraId})?`
- This ensures a position for an address is returned by a single app.
- In case a duplicate is detected, it is skipped and a warning is left in the log.
- For contract positions on the same address (like Uniswap V3 positions on a pool at different ranges), `extraId` can be specified.

Previously, positions with duplicate addresses were allowed but they were all returning the same first resolved position for it.
This should solve the issues discussed in https://github.com/valora-inc/hooks/pull/433

Fixes RET-1110


gnardini (approved):
Looks great! 😄

satish-ravi (approved):

PR #453, opened 2024-07-16
refactor: unified config access and simplifications for better devx

jeanregisser:
## Main Changes
1. Config moved outside of the API folder
2. Simplification of `getPositions` function
3. Consistency in config values
4. Improved developer experience

### 1. Config moved outside of the API folder
The config is now more universally accessible and not limited to the API. The `getClient` function, located outside of the API folder, was already accessing it from the runtime.

### 2. Simplification of `getPositions` function
Removed the need to pass `GET_TOKEN_INFO_URL` to `getPositions`, simplifying its usage.

### 3. Consistency in config values
There was a mismatch with the `getPositions` script, which was passing a default value for `GET_TOKEN_INFO_URL` that differed from the defaults set in `production.yaml` and `staging.yaml`. This caused the $ price to be unavailable during development. Now, all components read the same config, ensuring consistency and proper defaults for development.

### 4. Improved developer experience
With these updates, running `yarn start` or `yarn getPositions ...` works out-of-the-box after cloning the repo, providing a smoother experience for developers.

### DIFF
--- a/src/runtime/client.ts
+++ b/src/runtime/client.ts
@@ -10,7 +10,7 @@ import {
   sepolia,
 } from 'viem/chains'
 import { NetworkId } from '../types/networkId'
-import { getConfig } from '../api/config'
+import { getConfig } from '../config'
 
 const networkIdToViemChain: Record<NetworkId, Chain> = {
   [NetworkId['celo-mainnet']]: celo,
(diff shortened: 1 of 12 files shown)


src/runtime/client.ts
```diff
@@ -10,7 +10,7 @@ import {
   sepolia,
 } from 'viem/chains'
 import { NetworkId } from '../types/networkId'
-import { getConfig } from '../api/config'
```
jeanregisser, line 13: This is why I wanted the config out of the API folder.

satish-ravi (approved):

PR #455, opened 2024-07-18
feat: support polygon, base and their testnet

jeanregisser:
This is a quick change to unblock all requests failing since the wallet is now doing requests with `base` as `networkId`.

I'll update the test account to have a few positions too on these networks and will update the e2e tests separately.

### DIFF
--- a/src/apps/aave/positions.ts
+++ b/src/apps/aave/positions.ts
@@ -48,6 +48,19 @@ const AAVE_V3_ADDRESSES_BY_NETWORK_ID: Record<
     poolAddressesProvider: '0x36616cf17557639614c1cddb356b1b83fc0b2132',
     uiPoolDataProvider: '0x86e2938dae289763d4e09a7e42c5ccca62cf9809',
   },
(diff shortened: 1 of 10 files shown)


src/apps/aave/positions.ts
```diff
@@ -48,6 +48,19 @@ const AAVE_V3_ADDRESSES_BY_NETWORK_ID: Record<
     poolAddressesProvider: '0x36616cf17557639614c1cddb356b1b83fc0b2132',
     uiPoolDataProvider: '0x86e2938dae289763d4e09a7e42c5ccca62cf9809',
   },
+  [NetworkId['polygon-pos-mainnet']]: {
+    poolAddressesProvider: '0xa97684ead0e402dc232d5a977953df7ecbab3cdb',
+    uiPoolDataProvider: '0xc69728f11e9e6127733751c8410432913123acf1',
+  },
+  [NetworkId['polygon-pos-amoy']]: undefined,
```
jeanregisser, line 55: They only have Mumbai, which is deprecated now.

silasbw (approved):

PR #472, opened 2024-07-30
feat: add allbridge withdraw hook

MuckT:
### Description
- Adds a withdraw shortcut for AllBridge pools.

### Testing Plan
- [x] Test with Valora locally
- [x] Add Unit tests
- [x] [Successful withdrawal](https://celoscan.io/tx/0x5cdd4694a360637452d5038d41ccb1116f0dbada7d28fc0756f000d13e5989f9) with hooks cli: 

```
yarn triggerShortcut --networkId celo-mainnet --address 0x5ae3567405a8b743fd25cec7a435cca42277318e --positionAddress 0xfb2c7c10e731ebe96dabdf4a96d656bfe8e2b5af --app allbridge --shortcut withdraw --token.amount 1 --token.decimals 3 --mnemonic "<mnemonic>" --derivationPath "<derivationPath>"
```


src/apps/allbridge/shortcuts.ts
```diff
@@ -85,6 +85,40 @@ const hook: ShortcutsHook = {
...
+            decimals: z.coerce.number(),
+            amount: z.string(),
+          }),
+          positionAddress: ZodAddressLowerCased,
+        },
+        async onTrigger({ networkId, address, token, positionAddress }) {
+          const walletAddress = address as Address
+          const transactions: Transaction[] = []
+
+          const amountToWithdraw = parseUnits(token.amount, token.decimals)
```
jeanregisser, line 104: Just noting: the logic here will have to match what `withdrawTokenId` we set on the position.

https://github.com/valora-inc/hooks/blob/bc9d1b522ca9ba09d4d94f0af14a8c4619ebc60a/src/types/positions.ts#L68

Users will enter in fiat amount probably?

cc @finnian0826 @satish-ravi

jeanregisser (approved):
🚀

src/apps/allbridge/shortcuts.ts
finnian0826: @jeanregisser I think the current plan is to not allow withdrawing custom amounts so we would just pass the full token amount, but even if amount is entered in fiat we can easily convert to token amount in the wallet right?

PR #473, opened 2024-07-31
feat(pools): Update dataProps schema

finnian0826:
Discussion in slack thread [here](https://valora-app.slack.com/archives/C04NFRMSZQX/p1722374757544289).

Follow up [issue](https://linear.app/valora/issue/ACT-1328/placeholder-add-manageurl-and-contractcreatedat-to-aave-dataprops) to add manageUrl and contractCreatedAt to Aave dataProps (optional fields, will be used in UI).

Follow up [issue](https://linear.app/valora/issue/ACT-1329/placeholder-add-earnings-translation-strings-to-wallet) to add translation strings to wallet.

Command:
```
yarn getPositions --networkId arbitrum-one --apps aave --address 0x644b896b0c45717215c0f10501b20fafd9bceb24
```
Returned dataProps:
```
"dataProps": {
   "yieldRates": [
    {
     "yieldRatePercentage": 6.5327629537484055,
     "label": "earnFlow.yieldRateLabels.earningsApy",
     "tokenId": "arbitrum-one:0xaf88d065e77c8cc2239327c5edb3a432268e5831"
    }
   ],
   "earningItems": [],
   "depositTokenId": "arbitrum-one:0xaf88d065e77c8cc2239327c5edb3a432268e5831",
   "withdrawTokenId": "arbitrum-one:0x724dc807b04555b71ed48a6896b6f41593b8c637"
  },
```
Note: earningItems is empty because there are no longer rewards for this pool, if there is an address that has rewards I can test with that to ensure that field is correct


src/types/positions.ts
```diff
@@ -62,8 +62,24 @@ export interface DataPropsContext {
 // But in the future we could parameterize this
 export type DataProps = EarnDataProps
 
+export interface YieldRate {
+  yieldRatePercentage: number
```
satish-ravi, line 66: nit: `percentage`

src/apps/aave/positions.ts
```diff
@@ -121,7 +121,30 @@ const hook: PositionsHook = {
...
+                  }),
+                },
+              ],
+              earningItems: aTokenRewardsInfo?.length
+                ? aTokenRewardsInfo.map((info) => ({
+                    amount: toDecimalNumber(
+                      info.userUnclaimedRewards,
+                      info.rewardTokenDecimals,
+                    ),
+                    label: 'earnFlow.earningItemLabels.rewards',
```
satish-ravi, line 141: This seems tightly coupled with the wallet, wonder if we should define some kind of enum, and then the wallet can map this to an appropriate string. Although that would mean there would be BC issues if we want to add new types. @jeanregisser what do you think?

jeanregisser: Yes I think eventually the hooks repo/service should be localized.
I left a comment about this in https://linear.app/valora/issue/ACT-1329/add-earnings-hooks-translation-strings-to-wallet#comment-8c01cfac

jeanregisser (approved):
💯

PR #474, opened 2024-08-01
feat: optimize RPC calls by enabling client batching via multicall

jeanregisser:
I noticed getting positions without a user address could be quite slow or return 429 errors.
This is because the runtime was doing many more RPC calls in that case compared to when a user address is provided.

Specifically, getting earn positions was problematic on Arbitrum.
Or getting all curve positions.

On my local machine, running `yarn getPositions --networkId arbitrum-one --apps aave` would easily take **~40 secs** (and often was returning 429 with the defaut Arbitrum RPC).
Now with this change, it takes **~4** secs, and no more 429!
Yeah **10x** speed improvement! 😄 

Note: in some hooks, we could benefit from switching to implicit multicall via batching too. Because we fetch definitions in [parallel](https://github.com/valora-inc/hooks/blob/a8c099d0734f9b3ba285c638c354f1a87e9e5141/src/runtime/getPositions.ts#L406-L408), if the hook itself would call `readContract` (or via `Promise.all`), all these async calls would be batched too.
We can try this further optimization in the future.


src/runtime/getPositions.ts
```diff
@@ -238,32 +265,6 @@ async function resolveAppTokenPosition(
...
-    allowFailure: false,
-  })
-
-  let balance = contractData[0]
-  const totalSupply = contractData[1]
-  // If no user address is provided, use 0
-  if (!address) {
-    balance = 0n
-  }
-
```
jeanregisser, line 266: This was moved out of this function, directly when fetching tokensInfo, this way it's much faster cause it benefits from the client batching.

src/runtime/getPositions.ts
```diff
@@ -454,27 +454,40 @@ export async function getPositions(
...
+        baseTokensInfo[definitionTokenId] ||
+        unlistedBaseTokensInfo[definitionTokenId] ||
+        visitedDefinitions[definitionTokenId]?.type ===
+          'app-token-definition' ||
+        unresolvedTokenDefinitionsByTokenId[definitionTokenId]
+      ) {
+        continue
+      }
+      unresolvedTokenDefinitionsByTokenId[definitionTokenId] = tokenDefinition
+    }
```
jeanregisser, line 482: Updated the logic so we deduplicate the token definition.
which ends up saving on RPC calls too.

src/runtime/getPositions.ts
```diff
@@ -71,19 +73,23 @@ async function getBaseTokensInfo(
...
     }
   }
   return tokensInfo
 }
 
 async function getERC20TokenInfo(
-  address: Address,
+  contractAddress: Address,
   networkId: NetworkId,
+  address?: Address,
```
bakoushin, line 87: i had to read through all the function code to understand what is this `address`. perhaps `userAddress` may be more meaningful and could've saved me from that 😅

PR #478, opened 2024-08-02
feat(earn): Add allbridge pool to getEarnPositions

finnian0826:
Update getEarnPositions to include the [Allbridge Celo USDT pool](https://core.allbridge.io/pools?chain=CEL).

**Testing:**
Unit test added


src/api/index.ts
```diff
@@ -21,11 +21,13 @@ import { Transaction } from '../types/shortcuts'
...
   NetworkId['arbitrum-one'],
   NetworkId['arbitrum-sepolia'],
+  NetworkId['celo-mainnet'],
 ])
-const EARN_SUPPORTED_APP_IDS = ['aave']
+const EARN_SUPPORTED_APP_IDS = ['aave', 'allbridge']
 const EARN_SUPPORTED_POSITION_IDS = new Set([
   `${NetworkId['arbitrum-one']}:0x724dc807b04555b71ed48a6896b6f41593b8c637`,
   `${NetworkId['arbitrum-sepolia']}:0x460b97bd498e1157530aeb3086301d5225b91216`,
+  `${NetworkId['celo-mainnet']}:0xfb2C7c10e731EBe96Dabdf4A96D656Bfe8e2b5Af`,
```
satish-ravi, line 30: is there a testnet version we can add here similar to aave? Also, would be good to leave a comment on what pool this is, doesn't seem obvious with just a list of position ids

finnian0826: I wasn't sure, on the Allbridge Celo pools [page](https://core.allbridge.io/pools?chain=CEL) I didn't see anywhere to find a testnet version, and there aren't any docs. Is there a way given a [celoscan mainnet contract](https://celoscan.io/address/0xfb2C7c10e731EBe96Dabdf4A96D656Bfe8e2b5Af) to easily find it's testnet counterpart (if it exists)?

finnian0826: Added a comment in [571d07a](https://github.com/valora-inc/hooks/pull/478/commits/571d07a2f0d30491c067fb880687a6b643636539), not sure if there is a cleaner way, I just labeled each pool with provider and deposit token

satish-ravi: hmm if its not readily available, its fine to skip the testnet one. And the comment is good. Just want some easy way to identify by protocol, as it can get lost when this list gets big

satish-ravi (approved):

PR #496, opened 2024-08-13
fix(getEarnPositions): Only support mainnet networkIds in production and testnet networkIds in staging

jh2oman:
This change will make it easier for blockchain api to get all available earn positions.


finnian0826 (approved):
lgtm, should we also do something similar for `EARN_SUPPORTED_POSITION_IDS` since those are networkId dependent? having the arbitrum-sepolia pool in that list when on mainnet shouldn't be an issue, just a way to cleanup

src/config/index.ts
```diff
@@ -64,6 +68,12 @@ export function getConfig(): Config {
       .string()
       .default('')
       .transform((val) => (val === '' ? [] : val.split(','))),
+    EARN_SUPPORTED_NETWORK_IDS: z
+      .string()
+      .default('')
+      .transform((val) =>
+        val === '' ? ([] as NetworkId[]) : (val.split(',') as NetworkId[]),
```
finnian0826, line 75: Is there a reason for the empty string check here but not in prod?

jh2oman: here the string can be optional and default to empty string. in prod its not optional so we don't need to handle the empty case

jh2oman:
@finnian0826 I think that `EARN_SUPPORTED_POSITION_IDS` is fine where it is actually. I think of that as more some constants as opposed to per environment config. And now that the supported networks are per-environment there is no way that users could accidentally access sepolia from mainnet or vice versa.

PR #501, opened 2024-08-19
feat: Add string localization to hooks

finnian0826:
Mostly copied from `dapp-list` repo, also updated function calls to use object params

TODO:
- Update wallet `getPositions` and `getEarnPositions` API calls to pass language in the `Accept-Language` header
- Sync integration with crowdin

Made follow up ticket to:
-  [Update `displayProps`](https://linear.app/valora/issue/ACT-1350/placeholder-update-displayprops-and-possibly-other-places-in-hooks-to) and other places hardcoded strings are used to use translations instead

Testing (label fields generated from translations file):
```
yarn getPositions --networkId celo-mainnet --apps allbridge --address 0x644b896b0c45717215c0f10501b20fafd9bceb24
```
```
 "dataProps": {
   ...
   "yieldRates": [
    {
     "percentage": 0.5118663668900737,
     "label": "Earnings APR",
     "tokenId": "celo-mainnet:0x48065fbbe25f71c9282ddf5e1cd6d6a887483d5e"
    }
   ],
   "earningItems": [
    {
     "amount": "0.001047",
     "label": "Earnings",
     "tokenId": "celo-mainnet:0x48065fbbe25f71c9282ddf5e1cd6d6a887483d5e"
    }
   ],
   ...
  },
```


src/api/index.ts
```diff
@@ -27,6 +30,18 @@ const EARN_SUPPORTED_POSITION_IDS = new Set([
...
 
+i18next
+  .use(Backend)
+  .use(i18nextMiddleware.LanguageDetector)
+  .init({
+    backend: {
+      // eslint-disable-next-line no-path-concat
+      loadPath: '../../locales/{{lng}}.json',
+    },
+    fallbackLng: 'en',
```
finnian0826, line 41: it looks like english is fallback language for the wallet, should we do the same here?

src/api/index.ts
```diff
@@ -27,6 +30,18 @@ const EARN_SUPPORTED_POSITION_IDS = new Set([
...
+    preload: ['en'],
```
finnian0826, line 42: Wasn't sure which languages to preload, can add more if we want.

src/types/positions.ts
```diff
@@ -11,6 +12,7 @@ export interface PositionsHook {
   getPositionDefinitions(
     networkId: NetworkId,
     address?: string,
+    t?: TFunction<'translation', undefined>,
```
finnian0826, line 15: Wanted to make this required, but there are a lot of tests that use it, wasn't sure how to mock this function for a test (typescript was yelling at me when using jest.fn()), and there were a lot of places it would have to be added.

src/runtime/getPositions.ts
```diff
@@ -392,17 +393,23 @@ function addSourceAppId<T>(definition: T, sourceAppId: string) {
...
+export async function getPositions({
+  networkId,
+  address,
+  appIds = [],
+  t,
+}: {
+  networkId: NetworkId
+  address: string | undefined
+  appIds: string[]
+  t?: TFunction<'translation', undefined>
```
finnian0826, line 405: Also wanted to make it required here, but again a lot of tests use it and wasn't sure how to mock well.

jeanregisser: ah yes, I propose we let actual i18n happen.
See https://github.com/valora-inc/hooks/pull/505

src/types/positions.ts
jeanregisser: I propose we follow what I did in https://github.com/valora-inc/hooks/pull/505
and apply it here too.

jeanregisser (commented):
Looks great!! 🚀

src/api/index.ts
```diff
@@ -27,6 +30,23 @@ const EARN_SUPPORTED_POSITION_IDS = new Set([
...
+
+i18next
+  .use(Backend)
+  .use(i18nextMiddleware.LanguageDetector)
+  .init({
+    backend: {
+      // eslint-disable-next-line no-path-concat
+      loadPath: '../../locales/{{lng}}.json',
+    },
+    fallbackLng: DEFAULT_LANGUAGE,
```
jeanregisser, line 43: Sorry I think fallbackLng should be `base` since we don't have `en.json` yet,

src/apps/allbridge/positions.ts
```diff
@@ -126,7 +126,7 @@ const hook: PositionsHook = {
               yieldRates: [
                 {
                   percentage: apr,
-                  label: 'Earnings APR', // TODO(ACT-1331): Replace with localized string
+                  label: t ? t('yieldRates.earningsApr') : 'Earnings APR',
```
jeanregisser, line 129:
```suggestion
                  label: t('yieldRates.earningsApr'),
```

src/apps/allbridge/positions.ts
```diff
@@ -140,7 +140,7 @@ const hook: PositionsHook = {
                         pendingReward,
                         tokenInfo.decimals,
                       ),
-                      label: 'Earnings', // TODO(ACT-1331): Replace with localized string
+                      label: t ? t('earningItems.earnings') : 'Earnings',
```
jeanregisser, line 143:
```suggestion
                      label: t('earningItems.earnings'),
```

jeanregisser (approved):
Awesome! 🚀 

Already approved with some small suggestions.

PR #509, opened 2024-08-24
feat: add terms url to earn data props

satish-ravi:
Part of ACT-1342


src/types/positions.ts
```diff
@@ -84,6 +84,7 @@ export interface EarningItem {
 export interface EarnDataProps {
   contractCreatedAt?: string // ISO string
   manageUrl?: string
+  termsUrl?: string
```
satish-ravi, line 87: kept this optional to match other fields, but how would we handle the cases where these fields are missing in the wallet?  cc @finnian0826 @jeanregisser

MuckT (approved):

PR #513, opened 2024-08-27
feat: optionally simulate shortcut txs for aave

silasbw:


src/apps/aave/shortcuts.ts
```diff
@@ -83,14 +88,29 @@ const hook: ShortcutsHook = {
...
             }),
-            // TODO: consider moving this concern to the runtime
-            // which would simulate the transaction(s) to get these
-            gas: GAS,
-            estimatedGasUse: GAS / 3n,
           }
 
           transactions.push(supplyTx)
 
+          // TODO: consider moving this concern to the runtime
```
silasbw, line 95: This felt best as an end-to-end test, but I wasn't sure if/how we wanted to do that.  @jeanregisser ?

jeanregisser (approved):
🚀

PR #514, opened 2024-08-27
feat: consistent args for shortcuts

satish-ravi:
For clients to call deposit / withdraw / claim shorts for different providers, we need consistent input parameters for different providers. This updates position definitions to define a shortcutTriggerArgs that can be passed as is to `triggerShortcut`. Additionally, enforces `amount` on deposit / withdraw shortcuts as this needs to come from the client.

### DIFF
--- a/src/apps/aave/shortcuts.ts
+++ b/src/apps/aave/shortcuts.ts
@@ -32,22 +32,22 @@ const hook: ShortcutsHook = {
         name: 'Deposit',
         description: 'Lend your assets to earn interest',
         networkIds: [networkId],
-        // category: 'deposit',
+        category: 'deposit',
         triggerInputShape: {
           token: z.object({
             // TODO: consider requiring only tokenId and (decimal) amount
             // Right now it would mean more changes in hooks
             address: ZodAddressLowerCased,
             decimals: z.coerce.number(),
-            amount: z.string(), // in decimal string
           }),
+          amount: z.string(), // in decimal string
         },
-        async onTrigger({ networkId, address, token }) {
+        async onTrigger({ networkId, address, token, amount }) {
           const walletAddress = address as Address
           const transactions: Transaction[] = []
 
           // amount in smallest unit
-          const amountToSupply = parseUnits(token.amount, token.decimals)
+          const amountToSupply = parseUnits(amount, token.decimals)
 
           const client = getClient(networkId)
 
@@ -99,7 +99,7 @@ const hook: ShortcutsHook = {
         name: 'Withdraw',
         description: 'Withdraw your assets',
         networkIds: [networkId],
-        // category: 'withdraw',
+        category: 'withdraw',
         triggerInputShape: {
           // This is the A token
           token: z.object({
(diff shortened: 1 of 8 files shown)


src/apps/aave/shortcuts.ts
```diff
@@ -32,22 +32,22 @@ const hook: ShortcutsHook = {
...
+        category: 'deposit',
         triggerInputShape: {
           token: z.object({
             // TODO: consider requiring only tokenId and (decimal) amount
             // Right now it would mean more changes in hooks
             address: ZodAddressLowerCased,
             decimals: z.coerce.number(),
-            amount: z.string(), // in decimal string
           }),
+          amount: z.string(), // in decimal string
```
jeanregisser, line 43: Could we leave this within the `token`?
Anticipating the case where we want to deposit multiple tokens.
For instance to deposit into a standard pool with 2 assets.

src/types/shortcuts.ts
```diff
@@ -1,25 +1,34 @@
...
   getShortcutDefinitions(
     networkId: NetworkId,
     address?: string,
-  ): Promise<ShortcutDefinition<any>[]>
+  ): Promise<ShortcutDefinition<ShortcutCategory, any>[]>
 }
 
-export interface ShortcutDefinition<TriggerInputShape extends ZodRawShape> {
+type TriggerInputShape<Category> = Category extends 'deposit' | 'withdraw'
+  ? ZodRawShape & { amount: z.ZodString }
```
jeanregisser, line 14: I feel it's slightly surprising only amount is required here.
An amount without a tokenId or address is not very useful.

Is there a problem adding the other required fields we want for `deposit` and `withdraw`?

src/apps/allbridge/positions.ts
```diff
@@ -107,6 +107,21 @@ const hook: PositionsHook = {
...
+                },
+                positionAddress: tokenInfo.poolAddress.toLowerCase(),
+              },
+              withdraw: {
+                token: {
+                  decimals: tokenInfo.decimals,
+                },
+                positionAddress: tokenInfo.poolAddress.toLowerCase(),
+              },
+            },
```
jeanregisser, line 124: I'm thinking maybe we use `shortcutTriggerArgs` to add only extra things the client doesn't already pass?

The client needs to be aware of the `deposit` and `withdraw` shortcut types and that it needs to pass token info an amounts.

jeanregisser (commented):
Great!

src/types/shortcuts.ts
satish-ravi: the reason I left it out was because it was the only thing that needs to be set by the client and everything else can come from shortcutTriggerArgs, but I agree that it's cleaner with having token id / address + amount together. 

I wonder if we should just do the input as an array of tokenIds + decimal amount? If we do a single token here and if we keep the address + decimals format right now, changing to tokenId would be tricky to maintain backwards compatibility. Guess another option to not require decimals is to directly take the amount in smallest unit, but we still need token address, which we have to lookup using the tokenId, at which point we can get decimals too. So probably fine to just require tokenId + decimal amount and then do the lookup. What do you think?

satish-ravi: Chatted offline and landed on a tokens format with tokenId and amount. As a temp workaround, address and decimals are passed via shortcutTriggerArgs, until we can look it up using tokenId

jeanregisser (approved):
🚀

PR #515, opened 2024-08-27
chore: tweak simulation ergonomics

silasbw:
This new approach handles cases where networks aren't supported and when simulation fails for some reason.

### DIFF
--- a/src/apps/aave/shortcuts.ts
+++ b/src/apps/aave/shortcuts.ts
@@ -12,8 +12,6 @@ import { poolV3Abi } from './abis/pool-v3'
 import { aTokenAbi } from './abis/atoken'
 import { AAVE_V3_ADDRESSES_BY_NETWORK_ID } from './constants'
 import { incentivesControllerV3Abi } from './abis/incentives-controller-v3'
-
-import { getConfig } from '../../config'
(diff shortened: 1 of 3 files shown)


silasbw:
I'll fix tests if folks prefer the approach in the PR over the existing one.

jeanregisser (commented):
Yeah this looks good!

One question, are we later planning to move this out of the shortcut `onTrigger` implementation?

The code calling `onTrigger` could take care of the simulation, without the hook definition calling it explicitly.
Maybe we'd add a `fallbackGas` that hooks authors could define.

silasbw:
Yes, the plan is to move it out of the `onTrigger` implementation eventually. I was unsure about the assumptions we'd make/contract we'd have for doing so.

For example: https://github.com/valora-inc/hooks/blob/main/src/apps/aave/shortcuts.ts#L104-L109

What if there are more than two serialized transactions?  What if there's some combination of concurrent and serialized transactions?  What if I'm just overthinking?

jeanregisser:
I see, I'm fine trying our best guess here, directly outside of the shortcut definition.
It would work similarly for `N` TXs I think.

silasbw:
OK -- I can do that in a follow on!

jeanregisser (approved):

PR #517, opened 2024-08-27
chore: enable simulation for allbridge

silasbw:


src/apps/allbridge/shortcuts.ts
```diff
@@ -74,14 +77,28 @@ const hook: ShortcutsHook = {
...
+            })
+            const supplySimulatedTx =
+              simulatedTransactions[simulatedTransactions.length - 1]
+
+            supplyTx.gas =
+              BigInt(supplySimulatedTx.gasNeeded) +
+              SIMULATED_DEPOSIT_GAS_PADDING
+            supplyTx.estimatedGasUse = BigInt(supplySimulatedTx.gasUsed)
+          } catch (error) {
+            supplyTx.gas = GAS
```
satish-ravi, line 98: do we need a log here? if this is due to an unexpected error, seems like it would be useful for debugging

silasbw: yeah, probably.  the annoying thing is that there will always be an error on Celo (because no simulation support).

satish-ravi (approved):

PR #518, opened 2024-08-28
fix(aave): incorrect rewards

jeanregisser:
This fixes the incorrect display of rewards for Aave.

The `userRewardsInformation` returned for each a/v/sToken from `getUserReservesIncentivesData`
with `userUnclaimedRewards` can't be used as is and requires some additional calculations to get the total rewards for the user.

In particular if there are multiple Aave assets with incentives yielding a given reward token, the `userUnclaimedRewards` value will be the same across those assets.
And it doesn't include the accrued rewards for the assets currently being held.

The details of these calculations to get the total rewards can be seen in https://github.com/aave/aave-utilities/blob/446d9af6f14154771c0343538b59e2aeb7b38e47/packages/math-utils/src/formatters/incentive/calculate-all-user-incentives.ts#L31

For simplicity we can call `getAllUserRewards` which returns all rewards at once.
Small caveat, we loose the info about which a/v/sToken contributed to the reward.

This means we now have one single position for all Aave rewards instead of one per a/v/sToken (per network).

See also this Slack [thread](https://valora-app.slack.com/archives/C04NFRMSZQX/p1724715936061539).

| Before | After |
|-|-|
| image | image |

### DIFF
--- a/src/apps/aave/positions.ts
+++ b/src/apps/aave/positions.ts
@@ -86,7 +86,7 @@ const hook: PositionsHook = {
         : [undefined, [undefined, undefined]]
 
     // Note: Instead of calling `getAllUserRewards`, we could use reserveIncentiveData and userIncentivesData to get all user rewards
-    // but it requires some additional calculations to get the accrued rewards (linked to a/v/sToken held) and the unclaimedRewards.
+    // but it requires some additional calculations to get the accrued rewards (linked to a/v/sToken held) on top of the unclaimedRewards to get total rewards.
     // This is for simplicity here.
     // See https://github.com/aave/aave-utilities/blob/446d9af6f14154771c0343538b59e2aeb7b38e47/packages/math-utils/src/formatters/incentive/calculate-all-user-incentives.ts#L31
     const allUserRewards = await client.readContract({
--- a/src/apps/aave/positions.ts
+++ b/src/apps/aave/positions.ts
@@ -26,6 +26,7 @@ import {
 } from './constants'
 import { aTokenAbi } from './abis/atoken'
 import { incentivesControllerV3Abi } from './abis/incentives-controller-v3'
+import { getAaveTokensWithIncentives } from './getAaveTokensWithIncentives'
 
 const COMPOUND_PERIOD = 365 * 24 * 60 * 60 // 1 year in seconds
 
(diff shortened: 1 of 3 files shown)


src/apps/aave/positions.ts
```diff
@@ -111,211 +150,152 @@ const hook: PositionsHook = {
...
-                  label: t('yieldRates.earningsApy'),
-                  tokenId: getTokenId({
-                    networkId,
-                    address: reserveData.underlyingAsset.toLowerCase(),
-                  }),
+                  address: reserveData.underlyingAsset.toLowerCase(),
+                  networkId,
                 },
               ],
-              earningItems: aTokenRewardsInfo?.length
```
jeanregisser, line 183: We could/should calculate the accrued rewards here (not yet claimed).

So maybe we should bite the bullet and copy the calculation logic? 😅 
Instead of using `getAllUserRewards`.

src/apps/aave/positions.ts
```diff
@@ -111,211 +150,152 @@ const hook: PositionsHook = {
...
-                  getTokenId({
-                    networkId,
-                    address: reserveData.aTokenAddress.toLowerCase(),
-                  })
-                ],
-              tvl: toSerializedDecimalNumber(
-                toDecimalNumber(totalSupplies[i], lpTokenDecimals[i]),
-              ),
-              yieldRates: [
+        // Include a token if the user has a balance
```
jeanregisser, line 163: It's easier to hide the whitespaces so the diff is easier to read here.

Screenshot 2024-08-28 at 15 32 44

src/apps/aave/positions.ts
```diff
@@ -63,25 +64,63 @@ const hook: PositionsHook = {
...
+    // This is for simplicity here.
+    // See https://github.com/aave/aave-utilities/blob/446d9af6f14154771c0343538b59e2aeb7b38e47/packages/math-utils/src/formatters/incentive/calculate-all-user-incentives.ts#L31
+    const allUserRewards = await client.readContract({
+      address: aaveAddresses.incentivesController,
+      abi: incentivesControllerV3Abi,
+      functionName: 'getAllUserRewards',
+      args: [
+        // This builds the list of a/v/sToken address with incentives
+        // Because `getAllUserRewards` reverts if we pass an address with no incentives
+        reserveIncentiveData
```
jh2oman, line 99: nit: looks like this code can be shared with the similar mapping done in shortcuts.

jh2oman (approved):
Approved once tests are updated

src/apps/aave/positions.ts
```diff
@@ -111,211 +150,152 @@ const hook: PositionsHook = {
...
-          ({
-            type: 'contract-position-definition',
-            networkId,
-            address: reserveData.aTokenAddress.toLowerCase(),
-            extraId: 'supply-incentives',
-            tokens: aTokenRewardsInfo.map((info) => ({
-              address: info.rewardTokenAddress.toLowerCase(),
+                  },
+                ],
+                earningItems: [],
```
finnian0826, line 214: Are there not going to be rewards for Aave anymore? If so then this being empty makes sense but if there are rewards from the specific pool that can be pulled out that should be here. Also if there are pool specific rewards, I think there should be an item added to `yieldRates` that corresponds to the rewards APR (I think this was overlooked when yieldRates was first aded).

jeanregisser: There are still rewards, but to calculate the correct contribution from a given pool we need to NOT do the "shortcut" I've taken here for now.

It's what I was trying to explain in: https://github.com/valora-inc/hooks/pull/518#discussion_r1734684950

src/apps/aave/positions.ts
jeanregisser: ah yes, I was lazy, fixed 😄

src/apps/aave/positions.ts
jeanregisser: I guess we can address this separately. The value that was returned before my changes here, was incorrect anyway.

PR #527, opened 2024-09-03
fix(beefy): add fallbackPriceUsd to CLM positions

jeanregisser:
This also adds fallbackPriceUsd for CLM positions.
So Beefy CLM positions don't miss part of the value.

| Before | After |
|-|-|
| IMG_7091 | IMG_7092 |

Before the change, because we don't have a $ price for MIGGLES, the MIGGLES-WETH position $ value was only showing the WETH part.


bakoushin (approved):
LGTM! 🐮

silasbw (approved):
MIGGLES!

PR #529, opened 2024-09-03
feat: add reward position ids to earn data props

satish-ravi:
This way clients can identify what reward is associated with an earn position

### DIFF
--- a/src/types/positions.ts
+++ b/src/types/positions.ts
@@ -91,7 +91,7 @@ export interface EarnDataProps {
   earningItems: EarningItem[]
   depositTokenId: string
   withdrawTokenId: string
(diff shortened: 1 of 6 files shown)


src/types/positions.ts
```diff
@@ -91,6 +91,7 @@ export interface EarnDataProps {
   earningItems: EarningItem[]
   depositTokenId: string
   withdrawTokenId: string
+  rewardPositionIds?: string[]
```
jh2oman, line 94: nit: we've been using the plural `rewards` more often. so I suggest `rewardsPositionIds`

jh2oman (approved):

PR #539, opened 2024-09-11
feat(aave): add swap and deposit shortcut

satish-ravi:
Step 1 of ACT-1354, adds a swap-deposit shortcut for aave which uses squid post hook to deposit into an aave pool after swapping to the required deposit token. Will make follow ups to:
- Add the same shortcut for allbridge
- Return swap info for the frontend


package.json
```diff
@@ -37,6 +37,7 @@
     "deploy:production:http": "yarn deploy:production --trigger-http --allow-unauthenticated"
   },
   "dependencies": {
+    "@0xsquid/squid-types": "^0.1.104",
```
satish-ravi, line 40: This provides some type safety, but we could also drop this and hardcode enums

src/apps/aave/shortcuts.ts
```diff
@@ -216,6 +219,72 @@ const hook: ShortcutsHook = {
           ]
         },
       }),
+      createShortcut({
+        id: 'swap-deposit',
```
satish-ravi, line 223: open to suggestions for a better id / category for this

src/runtime/prepareSwapTransactions.ts
satish-ravi, line 1: Is this better suited for `utils/` over `runtime/`?

src/runtime/prepareSwapTransactions.ts
```diff
@@ -0,0 +1,200 @@
...
+            (simulatedGasPadding?.[index] ?? 0n)
+          ).toString(),
+        }
+      }),
+    }
+  } catch (error) {
+    if (!(error instanceof UnsupportedSimulateRequest)) {
+      logger.warn(error, 'Unexpected error during simulateTransactions')
+    }
+    // use default already set in the postHook, no changes needed
```
silasbw, line 100: nit: seems better to set the default here or comment on https://github.com/valora-inc/hooks/pull/539/files#diff-57288d8f4298dae89c637e7276d8497e62e59a1035d846326f93a8ff855d19cdR84 that it needs to the last statement in the try block.  otherwise, like, someone might add additional statements to the try block that update postHookWithSimulatedGas (so postHookWithSimulatedGas gets set, but then needs updates, but then we throw and don't use the default).

src/apps/aave/shortcuts.ts
```diff
@@ -216,6 +219,72 @@ const hook: ShortcutsHook = {
...
+            args: [tokenAddress, amount, walletAddress, 0],
+          })
+
+          return await prepareSwapTransactions({
+            networkId,
+            swapFromToken,
+            swapToAddress: tokenAddress,
+            walletAddress,
+            simulatedGasPadding: [0n, SIMULATED_DEPOSIT_GAS_PADDING],
+            postHook: {
```
silasbw, line 255: can you add a comment linking to the squid docs for this?  on first read I had questions that might be most easily handled w/ docs link.

silasbw (approved):

src/runtime/prepareSwapTransactions.ts
```diff
@@ -0,0 +1,200 @@
...
+  networkId,
+  walletAddress,
+  simulatedGasPadding,
+}: {
+  swapFromToken: z.infer<typeof tokenAmountWithMetadata>
+  postHook: Omit<
+    SquidHook,
+    'fundAmount' | 'fundToken' | 'provider' | 'logoURI' | 'calls'
+  > & { calls: EvmContractCall[] } // we don't support CosmosCall
+  swapToAddress: Address
```
jeanregisser, line 66: nit:
```suggestion
  swapToTokenAddress: Address
```

jeanregisser (commented):
Awesome work!! 🚀 

I'm running short of time to fully review everything now, but will continue tomorrow.

src/runtime/prepareSwapTransactions.ts
jeanregisser: I'd say `utils/`?

jh2oman (approved):
Lots of work, but well organized. very nice!

src/apps/aave/shortcuts.ts
satish-ravi: `zap-in/out` works, should we also change the input param from `swapFromToken` to `zapFromToken`?

src/utils/prepareSwapTransactions.test.ts
```diff
@@ -0,0 +1,288 @@
...
+
+const mockGotPostJson = jest.fn()
+const mockReadContract = jest.fn()
+
+jest.mock('got', () => ({
+  HTTPError: jest.requireActual('got').HTTPError,
+  post: jest.fn(() => ({
+    json: mockGotPostJson,
+  })),
+}))
```
jeanregisser, line 16: nit: I started using https://mswjs.io/ in this repo for network request mocks.
So tests are agnostic from the fetch client being used.

For now request mocks are all centralized in [test/server.ts](https://github.com/valora-inc/hooks/blob/main/test/server.ts), but I think we could probably refactor to override / provide custom responses per test.

We can address this separately though.

jeanregisser (approved):
Can't wait to see this live 🙌

src/apps/aave/shortcuts.ts
jh2oman: I like `swap-deposit`. Its concise and obvious from the name what its doing.

satish-ravi: brought this up in activation sync and people seemed to prefer `swap-deposit`, so keep it as is.

src/utils/prepareSwapTransactions.test.ts
satish-ravi: Ah, this is just for unit tests right? Happy to make a follow up later to adjust this. The got mocking is  pretty messy.

src/apps/aave/shortcuts.ts
jeanregisser: Ok that’s fair. One is directly understandable for people not familiar with crypto.
Whereas, zap is used in several crypto products like Beefy, [Yearn](https://andrecronje.gitbook.io/yearn-finance/products/zap), Zapper, [PancakeSwap](https://docs.pancakeswap.finance/products/pancakeswap-exchange/zap/zap-v2), [KyberSwap](https://docs.kyberswap.com/getting-started/foundational-topics/decentralized-finance/zap), [Krystal](https://x.com/krystaldefi/status/1780162409333657954) or [Symbiosis Finance](https://app.symbiosis.finance/zap). And may become the industry term for this.

PR #541, opened 2024-09-12
feat(allbridge): add swap and deposit shortcut

satish-ravi:


jeanregisser (approved):
Great! 🚀 

This got me thinking, since the transactions being done in shortcuts are getting more complex, maybe it could be worth it to simulate those (could be simple with our simulation API) in the e2e tests.
But we could check the outcome too (the right token balances deposited in the contract, etc), by forking the networking with Anvil. Though this is more involved of course.

satish-ravi:
Good idea. I guess we do get some simulation with swap but should we make a placeholder issue for these @jeanregisser ?

PR #555, opened 2024-09-19
fix(got): add timeouts to all requests

jh2oman:
by default got has no timeouts, adding 10s timeouts to all requests


jeanregisser (approved):
Cool, thanks! 🙌

PR #556, opened 2024-09-19
chore: cache allbridge HTTP responses to avoid rate limits

silasbw:


satish-ravi (approved):

MuckT (approved):

MuckT:
It looks like we primarily want a TTL cache which seems like this might not satisfy: https://github.com/isaacs/node-lru-cache?tab=readme-ov-file#storage-bounds-safety

> If you truly wish to use a cache that is bound only by TTL expiration, consider using a Map object, and calling setTimeout to delete entries when they expire. It will perform much better than an LRU cache.

silasbw:
I don't think we care about the storage bounds much since we're caching one item and there's little need to proactively evict it when the TTL expires.

PR #560, opened 2024-09-20
chore: add enableSwapFee to TriggerInputShape for swap-deposit

silasbw:


src/types/shortcuts.ts
```diff
@@ -31,12 +31,22 @@ export const tokenAmountWithMetadata = z.object({
   isNative: z.boolean(),
 })
 
+export const ZodEnableSwapFee = z
```
satish-ravi, line 34: this is a post request, can we just use zod booleans?

src/apps/allbridge/shortcuts.ts
```diff
@@ -192,6 +193,7 @@ const hook: ShortcutsHook = {
         triggerInputShape: {
           swapFromToken: tokenAmountWithMetadata,
           // set via shortcutTriggerArgs, the deposit token and position addresses
+          enableSwapFee: ZodEnableSwapFee,
```
satish-ravi, line 196: can we move this above the comment? This has to come from the wallet

src/types/shortcuts.ts
silasbw: I don't think so:

> Note — Boolean coercion with z.coerce.boolean() may not work how you expect. Any [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) value is coerced to true, and any [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) value is coerced to false.

silasbw:
this doesn't the behavior of anything yet @MuckT  --  I'm mostly trying to propose the parameter name and values so we can get the wallet change done.

src/types/shortcuts.ts
silasbw: doh, yes we can.

MuckT (approved):

PR #561, opened 2024-09-21
chore(earn): pass enableSwapFee to getSwapQuote

MuckT:
### Description

Pass `enableSwapFee` down to `getSwapQuote`.

### Related Issues

Part of ACT-1379

PR #570, opened 2024-09-23
fix(fees): pass enableSwapFee to prepareSwapTransactions

MuckT:
### Description

Pass enableSwapFee correctly: [related thread](https://github.com/valora-inc/hooks/pull/561#discussion_r1769461313).

### DIFF
--- a/src/apps/aave/shortcuts.ts
+++ b/src/apps/aave/shortcuts.ts
@@ -234,7 +234,13 @@ const hook: ShortcutsHook = {
           // set via shortcutTriggerArgs, the deposit token's address
           tokenAddress: ZodAddressLowerCased,
         },
-        async onTrigger({ swapFromToken, tokenAddress, address, networkId }) {
+        async onTrigger({
+          swapFromToken,
+          tokenAddress,
+          address,
+          networkId,
+          enableSwapFee,
+        }) {
           const walletAddress = address as Address
           // use a placeholder non zero amount so tx simulation can succeed.
           // squid postHook will replace this with the actual amount after swap.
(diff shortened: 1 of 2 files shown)


src/apps/aave/shortcuts.ts
```diff
@@ -234,7 +234,13 @@ const hook: ShortcutsHook = {
           // set via shortcutTriggerArgs, the deposit token's address
           tokenAddress: ZodAddressLowerCased,
         },
-        async onTrigger({ swapFromToken, tokenAddress, address, networkId }) {
+        async onTrigger({
+          swapFromToken,
+          tokenAddress,
+          address,
+          networkId,
+          enableSwapFee,
```
jeanregisser, line 242: Can we keep the original name `enableAppFee` we've used until now?

satish-ravi: hmm I thought about this when @silasbw  first proposed this here https://github.com/valora-inc/hooks/pull/560, I was on the fence because `enableSwapFee` adds some context vs a generic `enableAppFee`, as it involves both swap and deposit, but being consistent also is nice. Unfortunately changing now is a breaking change for the wallet, we could make it backwards compatible or delay a release if you feel strongly about this

jeanregisser: I see! We explicitly decided not to use "swap fee" to avoid the ambiguity with swap fee from liquidity pools.

Here's the full thread for context: https://valora-app.slack.com/archives/C029Z1QMD7B/p1714472447288399

Probably not worth delaying the current release, but maybe a backwards compatible update would be nice.

PR #571, opened 2024-09-24
chore(got): add hooks to capture timing info

satish-ravi:
- Increase timeout from 10 to 15s - noticed failures from getting swap quote or simulating transactions which sometimes take ~12s
- Add a `beforeError` hook that logs request details if there's a timeout - some timeout errors seem to be missing which request failed so this gives additional info
- Add an `afterResponse` hook that logs time taken by a request - helps debugging which api requests are slow.

### DIFF
--- a/src/utils/got.ts
+++ b/src/utils/got.ts
@@ -1,9 +1,47 @@
-import got, { ExtendOptions } from 'got'
+import got, { ExtendOptions, TimeoutError } from 'got'
+import { logger } from '../log'
 
 // https://github.com/sindresorhus/got/blob/HEAD/documentation/quick-start.md#options
 const options: ExtendOptions = {
   timeout: {
-    request: 10 * 1000,
+    request: 15 * 1000,
+  },
+  hooks: {
+    beforeError: [
+      (err) => {
+        if (err instanceof TimeoutError) {
+          logger.warn(
+            {
+              err,
+              request: {
+                url: err.request.requestUrl,
+                options: err.request.options,
+                timings: err.timings,
+              },
+            },
+            `Request timed out on ${err.request.options.method} ${err.request.requestUrl}`,
+          )
+        }
+        return err
+      },
+    ],
+    afterResponse: [
+      (response) => {
+        logger.debug(
+          {
+            response: {
+              statusCode: response.statusCode,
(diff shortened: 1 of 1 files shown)


src/utils/got.ts
```diff
@@ -1,9 +1,47 @@
...
+            },
+            `Request timed out on ${err.request.options.method} ${err.request.requestUrl}`,
+          )
+        }
+        return err
+      },
+    ],
+    afterResponse: [
+      (response) => {
+        logger.info(
```
satish-ravi, line 30: I first wanted to use `debug` but realized `src/runtime/getPositions.ts` uses logger.debug extensively which can get really noisy if we enable debug logging in prod. We could change the runtime logs to use `trace` and use debug here though, but didn't seem worth it

src/utils/got.ts
```diff
@@ -1,9 +1,47 @@
...
+              request: {
+                url: err.request.requestUrl,
+                options: err.request.options,
+                timings: err.timings,
+              },
+            },
+            `Request timed out on ${err.request.options.method} ${err.request.requestUrl}`,
+          )
+        }
+        return err
```
jeanregisser, line 25: Do you think we should extend our custom got error serialization in https://github.com/valora-inc/logging/blob/6258d46cef6b04ed4b2ba42bd7fe000c55aacd44/src/logging.ts#L240-L249 to include timings too? (url was already there from what I remember).

src/utils/got.ts
jeanregisser: ah right!
Alternatively, we could have different logging "domains" / "components" where we could adjust the verbosity.
We could do this with [bunyan's child](https://github.com/trentm/node-bunyan?tab=readme-ov-file#logchild) feature.

jeanregisser (approved):
Great!! 🚀 

I wonder if we should have update our swap and simulation functions to be gen2 and have always 1 instance running to avoid those pesky cold starts?

src/utils/got.ts
satish-ravi: yep that seems like a good idea

src/utils/got.ts
satish-ravi: ooh, I'll look into this in a follow up

PR #574, opened 2024-09-24
chore(got): only log response timings if its greater than 5s

satish-ravi:
The 5s is a bit arbitrary but from the logs, most requests take a max of 2-3 seconds, so adding this to capture the really slow ones and reduce noise in the logs. Also use debug child logger

### DIFF
--- a/src/utils/got.ts
+++ b/src/utils/got.ts
@@ -27,18 +27,23 @@ const options: ExtendOptions = {
     ],
     afterResponse: [
       (response) => {
-        logger.info(
-          {
-            response: {
-              statusCode: response.statusCode,
-              statusMessage: response.statusMessage,
-              url: response.request.requestUrl,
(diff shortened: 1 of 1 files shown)


src/utils/got.ts
```diff
@@ -27,18 +29,23 @@ const options: ExtendOptions = {
...
-          {
-            response: {
-              statusCode: response.statusCode,
-              statusMessage: response.statusMessage,
-              url: response.request.requestUrl,
-              timings: response.timings,
-              method: response.request.options.method,
+        if (
+          response.timings.phases.total &&
+          response.timings.phases.total > 5000
```
MuckT, line 34: super nit: 
```suggestion
          response.timings.phases.total > 5 * 1000
```

MuckT (approved):

PR #580, opened 2024-09-26
fix(beefy): try catch on userVaults map callback

MuckT:
### Description

An error was being thrown in [here](https://github.com/valora-inc/hooks/blob/main/src/apps/beefy/positions.ts#L55), so I've wrapped the map callback with a try-catch to prevent other pools from not being returned when the error throws.

### Related Issues
Part of ACT-1375


finnian0826 (approved):

jeanregisser:
What error was being thrown? I'm slightly worried the try / catch is a bandaid and we should fix the underlying issue instead.

PR #581, opened 2024-09-26
fix(positions): made address optional on TokenDefinition type

MuckT:
### Description

Some pools use tokens that don't have addresses such as ETH. This PR makes the `address` an optional field in the Beefy response.

### Related Issues

Part of ACT-1375

### Testing

`yarn getPositions -n base-mainnet  -a 0x5ae3567405a8b743fd25cec7a435cca42277318e`

#### Returns:

```
┌─────────┬─────────┬─────────────┬──────────────────────────────────────────────┬────────────────┬──────────────────────────┬─────────────┬─────────┬────────────┬─────────────────────────────────────────────┬──────────────────────┐
│ (index) │ appId   │ type        │ address                                      │ network        │ title                    │ priceUsd    │ balance │ balanceUsd │ breakdown                                   │ availableShortcutIds │
├─────────┼─────────┼─────────────┼──────────────────────────────────────────────┼────────────────┼──────────────────────────┼─────────────┼─────────┼────────────┼─────────────────────────────────────────────┼──────────────────────┤
│ 0       │ 'beefy' │ 'app-token' │ '0xacd23676d95f3314d5a7538c635ffdb6e66c4f9f' │ 'base-mainnet' │ 'SONNE-USDC vLP (Vault)' │ '762348.05' │ '0.00'  │ '1.61'     │ '0.00 vAMM-SONNE/USDC ($1.61) @ $177978.31' │ []                   │
│ 1       │ 'beefy' │ 'app-token' │ '0x62e5b9934dcb87618cfc74b222305d16c997e8c1' │ 'base-mainnet' │ 'ETH (Vault)'            │ '2736.11'   │ '0.00'  │ '7.08'     │ '0.00 ETH ($7.08) @ $2641.41'               │ []                   │
└─────────┴─────────┴─────────────┴──────────────────────────────────────────────┴────────────────┴──────────────────────────┴─────────────┴─────────┴────────────┴─────────────────────────────────────────────┴──────────────────────┘

```


finnian0826:
Do these positions still show fine in the wallet with the address undefined?

MuckT:
@finnian0826 

Running hooks and setting `HOOKS_API_URL_MAINNET=http://10.0.0.30:18000` in Valora returns the Pool.

jeanregisser (commented):
Thanks! I've been thinking some more about this case and I'm thinking we should use the `0xeee...` address for that case.

The main reason is because I think it's better to avoid developers unintentionally setting the address to `undefined` without realizing, causing the position to have the wrong token.

For instance, doing this:
```ts
{
  address: something.tokenAddress,
  networkId,
}
```

And that `something.tokenAddress` type including `undefined`.

Also I think we should update the beefy API response type, which is the root cause of this issue:

https://github.com/valora-inc/hooks/blob/0b60e9df0abd01453afc666a445707d2c5fc391c/src/apps/beefy/api.ts#L10

It should be changed to `Address | undefined`.

Note: we could export and use: https://github.com/valora-inc/hooks/blob/f4abca1499ac88ee0f63e998099e9029a8c86973/src/runtime/isNative.ts#L5

src/apps/beefy/positions.ts
```diff
@@ -276,6 +280,7 @@ const beefyGovVaultsPositions = async (
     functionName: 'getUserClmPools',
     args: [
       address,
+      // @ts-expect-error tokenAddress from beefy response is not always defined
```
MuckT, line 283: @jeanregisser with the changes tried a few different ways to fix this error and was unsuccessful.

jeanregisser: Could/should we filter these to only those with a `tokenAddress` set?

MuckT: Fixed in d47baf4220bf1278b42ebc120ec9a18cd8b8364f!

jeanregisser (approved):
👍

PR #592, opened 2024-09-27
feat(beefy): add earn data props

satish-ravi:
Part 1 of ACT-1369, will add shortcut info in a follow up

### DIFF
--- a/src/apps/beefy/positions.ts
+++ b/src/apps/beefy/positions.ts
@@ -17,9 +17,19 @@ import {
   beefyClmVaultsMulticallBytecode,
 } from './abis/beefy-clm-vaults-multicall'
 import { beefyV2AppMulticallAbi } from './abis/beefy-v2-app-multicall'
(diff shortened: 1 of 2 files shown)


src/apps/beefy/positions.ts
```diff
@@ -75,6 +99,36 @@ const beefyAppTokenDefinition = (
       const { decimals } = tokensByTokenId[tokenId]
       return [toDecimalNumber(BigInt(vault.pricePerFullShare), decimals)]
     },
+    dataProps: {
+      depositTokenId: getTokenId({
```
finnian0826, line 103: does this only work for beefy single token pools? or will vaults only have a single deposit token?

satish-ravi: I only validated against a handful single token pools for now, other multiple token vaults also seem to have a single `tokenAddress`, but I am not sure what that exactly means. Investigating them seems out of scope for now as we are only supporting single token pools

finnian0826 (approved):

PR #594, opened 2024-09-28
feat(beefy): return positions with 0 balance when address is not set

satish-ravi:
So this can be used to display earn positions. Actually returning beefy earn positions will be done in ACT-1367

### DIFF
--- a/src/apps/beefy/positions.ts
+++ b/src/apps/beefy/positions.ts
@@ -208,7 +208,7 @@ const beefyBaseVaultsPositions = async ({
   t,
 }: {
   networkId: NetworkId
-  address: Address
+  address?: Address
   vaults: BaseBeefyVault[]
   multicallAddress: Address
   prices: BeefyPrices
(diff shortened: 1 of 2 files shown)


src/apps/beefy/positions.ts
```diff
@@ -220,35 +220,41 @@ const beefyBaseVaultsPositions = async ({
...
+  }
 
   if (!userVaults.length) {
     return []
   }
 
   const clmVaults = userVaults.filter((vault) => vault.type === 'cowcentrated')
   const info =
-    clmVaults.length === 0
+    clmVaults.length === 0 || !address // earn positions can't include clm vaults for now
```
jeanregisser, line 257: What's blocking including CLM vaults?

jeanregisser (approved):
🚀

src/apps/beefy/positions.ts
satish-ravi: I believe they all have multiple base tokens and are returned as contract positions, which is currently not supported in the wallet

PR #595, opened 2024-09-28
feat(beefy): return shortcut info from positions

satish-ravi:


src/apps/beefy/positions.ts
```diff
@@ -246,6 +265,17 @@ const beefyBaseVaultsPositions = async ({
...
+  const earnTokenDecimals = await Promise.all(
+    userVaults.map((vault) =>
+      client.readContract({
+        address: vault.earnedTokenAddress,
+        abi: vaultAbi,
+        functionName: 'decimals',
+        args: [],
+      }),
+    ),
+  )
```
jeanregisser, line 277: This reminds me we I wanted to also expose a function for `shortcutTriggerArgs` so we could access the resolved tokens info (with decimals), so hooks wouldn't have query this data directly. Similar to https://github.com/valora-inc/hooks/blob/5d8c276fd589646b60ded7df021701aeb605c319/src/types/positions.ts#L102
This is more an optimization.

satish-ravi: oh good idea, I'll do this in a follow up

PR #598, opened 2024-09-30
feat(earn): Add supportedPools query param to getEarnPositions

finnian0826:
If query param not provided, default to legacy hardcoded list.

Unit tests updated/added.

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -23,7 +23,7 @@ import { Transaction } from '../types/shortcuts'
 import { parseRequest } from './parseRequest'
 
 const EARN_SUPPORTED_APP_IDS = ['aave', 'allbridge']
-const EARN_SUPPORTED_POSITION_IDS = new Set([
+const LEGACY_EARN_SUPPORTED_POSITION_IDS = new Set([
   // Aave USDC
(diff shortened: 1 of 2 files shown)


src/api/index.ts
```diff
@@ -23,7 +23,7 @@ import { Transaction } from '../types/shortcuts'
 import { parseRequest } from './parseRequest'
 
 const EARN_SUPPORTED_APP_IDS = ['aave', 'allbridge']
-const EARN_SUPPORTED_POSITION_IDS = new Set([
+const LEGACY_EARN_SUPPORTED_POSITION_IDS = new Set([
```
satish-ravi, line 26: nit: avoid legacy naming. Maybe `DEFAULT_`?

src/api/index.ts
```diff
@@ -165,6 +165,11 @@ function createApp() {
         .array(z.nativeEnum(NetworkId))
         .nonempty()
         .or(z.nativeEnum(NetworkId)), // singleton arrays sometimes serialize as single values
+      supportedPools: z
```
satish-ravi, line 168: should we also get app ids from the request? seems like it's not optimal to get all positions for supported apps and then filter down

satish-ravi (approved):

PR #599, opened 2024-09-30
refactor(positions): use context for shortcut trigger args

satish-ravi:
Saves a contract call to get decimals

### DIFF
--- a/src/runtime/getPositions.ts
+++ b/src/runtime/getPositions.ts
@@ -12,7 +12,6 @@ import {
   DisplayProps,
   Position,
   PositionDefinition,
-  RawTokenInfo,
   ShortcutTriggerArgs,
   Token,
   TokenInfo,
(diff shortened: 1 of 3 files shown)


src/runtime/getPositions.ts
```diff
@@ -342,6 +339,8 @@ async function resolveContractPosition(
     resolvedTokensByTokenId,
   )
   const dataProps = getDataProps(positionDefinition, resolvedTokensByTokenId)
+  const shortcutTriggerArgs =
+    getShortcutTriggerArgs(positionDefinition, {} as any) ?? {}
```
satish-ravi, line 343: TODO: address this

src/types/positions.ts
```diff
@@ -95,6 +95,15 @@ export interface EarnDataProps {
   // We'll add more fields here as needed
 }
 
+interface ShortcutTriggerArgsContext {
+  tokenInfo: TokenInfo
```
satish-ravi, line 99: @jeanregisser unlike data / displayProps, we can't use `resolvedTokensById` here because we need the info of the token we're currently resolving, and this is contained in this `tokenInfo` field. But this applies only for app tokens and not contract positions, any preferences on how to deal with this for contract positions?

jeanregisser: oh I see, I guess we could pass `tokensByTokenId` instead, which includes all app tokens info?

src/types/positions.ts
```diff
@@ -186,6 +191,26 @@ export interface BaseToken extends AbstractToken {
...
+  address?: string
+  name: string
+  symbol: string
+  decimals: number
+  imageUrl: string
+  tokenId: string
+  networkId: NetworkId
+  isNative?: boolean
+  priceUsd?: string
+}
```
jeanregisser, line 204: nit: could we move this one to be internal only to `getPositions.ts`? I didn't want to expose it yet.

jeanregisser (approved):
Great! 🚀

PR #602, opened 2024-10-02
feat: add manageUrl to displayProps

MuckT:
### Description

- Adds an optional `manageUrl` to `displayprops` which can be specified per pool.
- `manageUrl` was added for the following apps:
  - Beefy
  - Compound
  - Curve
  - Uniswap
  - Example

### Related Issues
Part of ACT-1384

### DIFF
--- a/src/apps/beefy/positions.ts
+++ b/src/apps/beefy/positions.ts
@@ -93,6 +93,7 @@ const beefyAppTokenDefinition = ({
         description: 'Vault',
         imageUrl:
           'https://raw.githubusercontent.com/valora-inc/dapp-list/main/assets/beefy.png',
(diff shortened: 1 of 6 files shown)


finnian0826 (commented):
lgtm, a little confused why the curve API gives a list of withdraw URLs, are there multiple ways to withdraw or is that always just a one item list?

finnian0826 (approved):

MuckT:
@finnian0826 Curve gives a few arrays of URLs: swap, deposit, withdraw. I picked withdraw as only users with a balance will see this in Valora and it seemed like the most appropriate action; Once at the URL users can toggle between swap, deposit, and withdraw actions in the Curve UI.

PR #603, opened 2024-10-02
refactor(positions): make one call to getTokensInfo per request

satish-ravi:
Currently makes one call per network, this should cut the load 5x


jeanregisser (approved):
Awesome! Thanks! 🚀

MuckT (approved):

PR #604, opened 2024-10-03
test(e2e): remove halo fi due to intermittent failures

MuckT:
### Description

Halo Fi is shutting down. Some of our e2e tests are now failing intermittently with HaloFi. This PR removes those tests.


satish-ravi (approved):

jeanregisser (commented):
Thanks! Let's even remove HaloFi, the API is down and won't come back AFAIK.
Or do you have another plan?

PR #605, opened 2024-10-04
style: make manageUrl required

MuckT:
### Description

Aave's ManageUrl was only used in dataProps and this PR adds it to displayProps.

### DIFF
--- a/src/apps/aave/positions.ts
+++ b/src/apps/aave/positions.ts
@@ -175,6 +175,7 @@ const hook: PositionsHook = {
             title: `Claimable rewards`,
             description: 'For supplying and borrowing',
             imageUrl: AAVE_LOGO,
+            manageUrl
           },
         } satisfies ContractPositionDefinition)
       : null
--- a/src/apps/aave/positions.ts
+++ b/src/apps/aave/positions.ts
@@ -175,7 +175,7 @@ const hook: PositionsHook = {
             title: `Claimable rewards`,
             description: 'For supplying and borrowing',
             imageUrl: AAVE_LOGO,
-            manageUrl
+            manageUrl,
           },
         } satisfies ContractPositionDefinition)
       : null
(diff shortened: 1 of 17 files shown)


src/apps/aave/positions.ts
```diff
@@ -175,6 +175,7 @@ const hook: PositionsHook = {
             title: `Claimable rewards`,
             description: 'For supplying and borrowing',
             imageUrl: AAVE_LOGO,
+            manageUrl
```
satish-ravi, line 178: I don't see this for allbridge too. I assume this is for making positions clickable right? Should this be a required field in the types then? Seems like this should be present for all positions. Or if some positions don't have a url we can link to, can we enforce it in TS so that this field is explicitly specified as `undefined` rather than keeping it optional?

MuckT: @satish-ravi Addressed in 6718123ef43643c2d3a7daf9893a17db9cc5e699! Hopefully this more explicit definition will result in the `manageUrl` being provided more often.

satish-ravi (commented):
I see many apps' urls are undefined, but they do have websites (hedgey, stake-dao, ubeswap, etc). Can we add urls for these?

MuckT:
@satish-ravi added a mangeUrls for all the apps except HaloFi, Locked-CELO, and Mento (Only swap page found).

satish-ravi (approved):

src/apps/allbridge/positions.ts
```diff
@@ -124,6 +124,7 @@ const hook: PositionsHook = {
             title: `${tokenInfo.symbol} supply incentives`,
             description: 'Rewards for supplying',
             imageUrl: ALLBRIDGE_LOGO,
+            manageUrl: `${ALLBRIDGE_POOLS_BASE_URL}?chain=${NETWORK_ID_TO_ALLBRIDGE_CHAIN[networkId]}`,
```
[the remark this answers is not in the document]
MuckT: Fixed in 0256f2d0f8593d4fa5718258708159a9c1e96f69!

PR #620, opened 2024-10-14
feat(earn): Add dailyYieldRatePercentage to beefy dataProps response

jh2oman:
Code based on: https://github.com/beefyfinance/beefy-v2/blob/main/src/features/data/actions/apy.ts#L46

https://linear.app/valora/issue/ACT-1406/daily-rate

### Tested

Ran the Beefy UI locally and console.logged the exact totalDaily value from the UI. Verified that it matched the calculation here.

### DIFF
--- a/src/apps/beefy/api.ts
+++ b/src/apps/beefy/api.ts
@@ -6,7 +6,7 @@ export type BeefyVault = {
   id: string
   name: string
   type: string // cowcentrated, gov
-  subType: string
+  subType?: string
   token: string
   tokenAddress: Address | undefined
   tokenDecimals: number
--- a/src/apps/beefy/api.ts
+++ b/src/apps/beefy/api.ts
@@ -6,6 +6,7 @@ export type BeefyVault = {
   id: string
   name: string
   type: string // cowcentrated, gov
+  subType: string
   token: string
   tokenAddress: Address | undefined
   tokenDecimals: number
@@ -137,10 +138,10 @@ export async function getBeefyPrices(
   }
 }
 
-export async function getApys() {
+export async function getApyBreakdown() {
   return got
-    .get(`https://api.beefy.finance/apy/`)
-    .json<Record<string, number | undefined>>()
+    .get(`https://api.beefy.finance/apy/breakdown/`)
+    .json<Record<string, Record<string, number>>>()
 }
 
 export async function getTvls(
--- a/src/apps/beefy/positions.ts
+++ b/src/apps/beefy/positions.ts
@@ -74,7 +74,7 @@ const beefyAppTokenDefinition = ({
     vault.tokenAddress?.toLowerCase() ??
     networkIdToNativeAssetAddress[networkId]
   const tvl = tvls[vault.id]
-  const apy = apyBreakdown[vault.id].totalApy ?? 0
+  const apy = apyBreakdown[vault.id].totalApy || 0
   return {
     type: 'app-token-definition',
     networkId,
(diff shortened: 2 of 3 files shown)


src/apps/beefy/positions.ts
```diff
@@ -74,7 +74,7 @@ const beefyAppTokenDefinition = ({
     vault.tokenAddress?.toLowerCase() ??
     networkIdToNativeAssetAddress[networkId]
   const tvl = tvls[vault.id]
-  const apy = apys[vault.id]
+  const apy = apyBreakdown[vault.id].totalApy
```
MuckT, line 77: I see five or so `*-earnings` vaults that do not have earnings defined do these need a fallback?

jh2oman: good idea! I can be a bit more defensive on a few things here

src/apps/beefy/positions.ts
```diff
@@ -160,6 +164,62 @@ const beefyAppTokenDefinition = ({
...
+  } else {
+    // "uncompound" apy to get daily apr
+    totalDaily = _yearlyToDaily(apyBreakdown.totalApy)
+  }
+
+  // Presence of rewardPoolApr indicates new api calc that has correct totals
+  // [Old gov pools had their apr in the vaultApr field]
+  if (vault.type === 'gov' && vault.subType === 'cowcentrated') {
+    // anything in 'rewardPoolApr' (i.e. not in 'rewardPoolTradingApr') is considered a soft-boost on the pool
+    const additionalApr = apyBreakdown.rewardPoolApr || 0
```
satish-ravi, line 203:
```suggestion
    const additionalApr = apyBreakdown.rewardPoolApr ?? 0
```

src/apps/beefy/api.ts
```diff
@@ -6,6 +6,7 @@ export type BeefyVault = {
   id: string
   name: string
   type: string // cowcentrated, gov
+  subType: string
```
satish-ravi, line 9: is there a set of values that this can contain? Maybe we can use specific types for `type` and `subType`.

src/apps/beefy/positions.ts
```diff
@@ -160,6 +164,62 @@ const beefyAppTokenDefinition = ({
...
+      }
+    }
+  } else {
+    // "uncompound" apy to get daily apr
+    totalDaily = _yearlyToDaily(apyBreakdown.totalApy)
+  }
+
+  // Presence of rewardPoolApr indicates new api calc that has correct totals
+  // [Old gov pools had their apr in the vaultApr field]
+  if (vault.type === 'gov' && vault.subType === 'cowcentrated') {
```
satish-ravi, line 201: vault.type itself can be `cowcentrated` right? could we document what a pool with `type=='gov'` and `subType=='cowncentrated'` is? Is that `Old gov pools`? if so can we make it explicit in the comment?

src/apps/beefy/positions.ts
```diff
@@ -160,6 +164,62 @@ const beefyAppTokenDefinition = ({
...
+  }
+
+  // TODO: handle merkl campaigns which are off-chain rewards
+  // https://github.com/beefyfinance/beefy-v2/blob/4413697f3d3cb5e090d8bb6958b621a673f0d739/src/features/data/actions/apy.ts#L132
+
+  return totalDaily * 100
+}
+
+const _yearlyToDaily = (apy: number) => {
+  const g = Math.pow(10, Math.log10(apy + 1) / 365) - 1
```
satish-ravi, line 214: nit: use a better name than `g`

src/types/positions.ts
```diff
@@ -100,6 +100,7 @@ export interface EarnDataProps {
   rewardsPositionIds?: string[]
   claimType?: ClaimType
   withdrawalIncludesClaim?: boolean
+  dailyYieldRatePercentage?: number
```
satish-ravi, line 103: can we document what goes in `yieldRates` and why we have a specific `dailyYieldRatePercentage`?

src/apps/beefy/api.ts
jh2oman: It appears there are, they are not all located in one place throughout the code in beefy. What would be the value in adding them in right now though?

src/apps/beefy/positions.ts
satish-ravi: Looks like you did the opposite change 😄

src/apps/beefy/positions.ts
satish-ravi: my comment (and the one about subtype values) was mainly because I was confused why something would have `type=gov` and `subtype=cowcentrated` when type can also be cowcentrated. And I also didn't find any vault like that in the vaults response for arbitrum. If this was just copied from beefy as is, probably fine to leave as is.

satish-ravi (approved):

PR #631, opened 2024-10-24
test(beefy): Add unit tests for getDailyYieldRatePercentage

finnian0826:
Unit tests added to cover:
- when there are breakdown components
- when there are no breakdown components
- when it needs to remove `rewardPoolApr`

### DIFF
--- a/src/apps/beefy/positions.test.ts
+++ b/src/apps/beefy/positions.test.ts
@@ -1,0 +1,47 @@
+import { BaseBeefyVault } from './api'
+import { getDailyYieldRatePercentage } from './positions'
+
+const apyBreakdownWithCorrectComponents = {
+  vaultApr: 0.1,
+  clmApr: 0.2,
+  tradingApr: 0.3,
+}
+
+const apyBreakdownWithIncorrectComponents = {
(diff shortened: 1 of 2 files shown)


src/apps/beefy/positions.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  ...apyBreakdownWithCorrectComponents,
+  rewardPoolApr: 0.4,
+}
+
+const vault: BaseBeefyVault = {
+  type: 'gov',
+  subType: 'cowcentrated',
+} as BaseBeefyVault
+
+describe('getDailyYieldRatePercentage', () => {
```
satish-ravi, line 25: can we add test cases for returning 0? It should be for when `undefined` and I also see a isNaN check, not sure when that can happen in practice

finnian0826: added two new tests to cover these cases in [3dff1ad](https://github.com/valora-inc/hooks/pull/631/commits/3dff1ad7b2287522c5413ca6c99f54f3aa884f3e)

satish-ravi (approved):
PR naming nit: use `test` over `chore`

PR #633, opened 2024-10-25
fix(gooddollar): tell existing users to re-verify after their verification expires

jeanregisser:
Follow up to https://github.com/valora-inc/hooks/pull/632

The verification status for the test account we were using expired.

We now keep showing the claimable position for existing users that were previously verified and they can do so on the dapp.

### DIFF
--- a/src/apps/gooddollar/positions.ts
+++ b/src/apps/gooddollar/positions.ts
@@ -49,12 +49,17 @@ const hook: PositionsHook = {
       abi: identityAbi,
     } as const
 
-    const [isVerified, currentDay, periodStart, claimAmount] =
+    const [isVerified, isNotNewUser, currentDay, periodStart, claimAmount] =
       await client.multicall({
         contracts: [
           {
             ...identityContract,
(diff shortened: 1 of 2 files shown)


src/apps/gooddollar/positions.ts
```diff
@@ -96,7 +102,8 @@ const hook: PositionsHook = {
       displayProps: {
         title: 'Daily UBI',
         description: !isVerified
-          ? 'Verify on the dapp to claim'
+          ? // Note: existing users need to re-verify after some time
+            'Re-verify on the dapp to claim'
```
satish-ravi, line 106: do these need to be localized eventually?

satish-ravi (approved):

satish-ravi:
When would this expire again next? Should we document some steps on what action needs to be taken when that happens?

jeanregisser:
> When would this expire again next? Should we document some steps on what action needs to be taken when that happens?

It won't, I didn't re-verify the test account. But the logic was updated to keep showing a position telling expired users to re-verify.

PR #638, opened 2024-10-30
test(beefy): Add unit tests for beefy positions hook

finnian0826:
Focused on unit tests that hit `beefyAppTokenDefinition`, since that is what the earn positions are. Unit tests added for:

- Confirming `getInfo` returns the expected fields for beefy
- `getPositionDefinitions` returns positions with balance when supported networkId and address passed in 
- `getPositionDefinitions` returns all positions when supported networkId passed in but no address
- `getPositionDefinitions` returns empty list for unsupported networkId


src/apps/beefy/positions.test.ts
```diff
@@ -17,6 +112,78 @@ const vault: BaseBeefyVault = {
...
+    withdrawTokenId: 'arbitrum-one:0x987654321',
+    yieldRates: [
+      {
+        label: 'yieldRates.earningsApy',
+        percentage: 31,
+        tokenId: 'arbitrum-one:0x123456789',
+      },
+    ],
+  },
+  displayProps: expect.any(Function),
```
finnian0826, line 137: `displayProps`, `pricePerShare`, and `shortcutTriggerArgs` are returned as functions for Beefy earn positions. Wasn't sure how to check for a specific function implementation so just used `expect.any(Function)` instead.

src/apps/beefy/positions.test.ts
```diff
@@ -1,5 +1,100 @@
...
+
+const mockBeefyVault1: BaseBeefyVault = {
+  id: 'vault1',
+  name: 'Vault 1',
+  type: 'auto',
+  subType: 'concentrated',
+  token: 'vault1Token',
+  tokenAddress: '0x123456789' as Address,
+  tokenDecimals: 8,
+  tokenProviderId: '',
```
finnian0826, line 32: These empty fields are not used by the implementation for Earn positions. I can add dummy values if wanted but wasn't exactly sure what to put.

src/apps/beefy/positions.test.ts
```diff
@@ -1,5 +1,100 @@
...
+  earnedToken: '',
+  earnContractAddress: '' as Address,
+  status: '',
+  platformId: '',
+  assets: [],
+  risks: [],
```
finnian0826, line 38: Explicit unit testing for safety score already exists so left that out of implementation here, but can add in if wanted.

src/apps/beefy/positions.test.ts
MuckT: Could you change this to have a default mock mockBeefyVault with these empty values? Then, in cases like `mockBeefyVault1` and `mockBeefyVault2`, destructure that object and add the appropriate new values. With all the values in each object hardcoded, it's hard for me to tell what's changed between them at a glance.

MuckT (approved):
Thanks for adding these! I wonder if it would have been easier to mock Beefy's response, but I think you'd still need a mock for `readContract,` so it's probably not worth it.

PR #648, opened 2024-11-14
feat: adjust estimated gas for deposit shortcuts

satish-ravi:
For ACT-1394. 
Changes include:
- Adjusting postHook default estimatedGas for swap-deposit txs
- Using gas from simulation instead of swap provider for swap-deposit txs 
  - Swaps use swap provider gas and ignores simulation results here for gas limit of the tx https://github.com/valora-inc/valora-rest-api/blob/ddf710202b63bfd9f02dc104b6aa812d64809c47/src/swaps/simulateSwap.ts#L156, this can be adjusted in swaps but for now trying this out in swap deposit
- Adding 15% buffer to simulation results to be extra cautious
  - Replaces the static 250k padding for deposits, but that was actually needed because of an issue with negative gas refunds (fixed in https://github.com/valora-inc/valora-rest-api/pull/1201) 

This can be monitored and adjusted based on these metrics: https://mixpanel.com/project/2786968/view/3321744/app/insights#vAou6TfFHjbW

### DIFF
--- a/src/utils/prepareSwapTransactions.test.ts
+++ b/src/utils/prepareSwapTransactions.test.ts
@@ -85,7 +85,6 @@ describe('prepareSwapTransactions', () => {
       swapFromToken: mockNativeSwapFromToken,
       swapToTokenAddress: '0x724dc807b04555b71ed48a6896b6f41593b8c637',
       postHook: mockPostHook,
-      simulatedGasPadding: [1n, 100n],
     })
 
     expect(transactions).toHaveLength(1)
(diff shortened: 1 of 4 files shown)


src/utils/prepareSwapTransactions.test.ts
```diff
@@ -206,50 +205,11 @@ describe('prepareSwapTransactions', () => {
       to: '0x12345678',
       data: '0xswapdata',
       value: 111n,
-      gas: 12345n,
+      gas: 14042n,
       estimatedGasUse: 12211n,
     })
   })
 
-  it('uses default gas for postHook', async () => {
```
satish-ravi, line 214: this test was redundant, post hooks always use the default gas and this is tested in other tests (left over cleanup from #582)

MuckT (approved):

PR #672, opened 2024-11-29
feat: add position for staked WCT

jeanregisser:


kathaypacific (approved):
should we refer to all of this as "reown"? i thought they were rebranding but the token is WCT which i assume means wallet connect token :/ so confusing...

jeanregisser:
They are more splitting than fully rebranding. WCT is part of the WalletConnect Network, developped by the WalletConnect Foundation from what I understand.

> The WalletConnect Foundation, alongside node operators and community contributors, will play a critical role in supporting the growth, security, and decentralization of the WalletConnect Network and the expansion of the WalletConnect ecosystem.
 
> WalletConnect Inc. is now Reown. The WalletConnect Network operates independently from Reown.

https://walletconnect.network/blog/introducing-the-walletconnect-foundation

PR #680, opened 2024-12-02
fix: valora app version check for Aave positions

jeanregisser:
I was randomly reading the code and noticed the check to include/exclude Aave positions would fail once we hit v `1.100`. Which will be the next release.

This fixes it, and also defaults to including Aave for other Mobile Stack apps (any other User-Agent).
Which is what I think we want.

### DIFF
--- a/src/api/index.test.ts
+++ b/src/api/index.test.ts
@@ -12,7 +12,7 @@ import { getConfig } from '../config'
 
 jest.mock('../config')
 jest.mocked(getConfig).mockReturnValue({
-  POSITION_IDS: [],
+  POSITION_IDS: ['uniswap', 'aave', 'curve'],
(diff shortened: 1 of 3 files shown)


MuckT (approved):
🚀

src/api/index.test.ts
```diff
@@ -322,6 +326,37 @@ describe('GET /getPositions', () => {
       data: TEST_POSITIONS_CELO,
     })
   })
+
+  for (const [userAgent, shouldIncludeAave] of [
```
satish-ravi, line 330: is the reason to use for loop over it.each is for better test names? or something else?

satish-ravi (approved):
Great catch!

src/api/index.test.ts
jeanregisser: Yes for better test name.

PR #692, opened 2024-12-12
feat(sommelier): add positions hook

kathaypacific:
Related to RET-1258


src/apps/sommelier/positions.test.ts
jeanregisser, line 1: Unit tests are nice.
Could you also add an e2e one using the shared test account? To have confidence it works for real.

src/apps/sommelier/positions.ts
```diff
@@ -0,0 +1,101 @@
...
+      name: 'Sommelier',
+      description: 'Sommelier vaults',
+    }
+  },
+  async getPositionDefinitions({ networkId }) {
+    const client = getClient(networkId)
+    const cellars = await getSommStrategiesData(networkId)
+    const positionDefinitions: PositionDefinition[] = []
+
+    const results = await Promise.allSettled(
```
jeanregisser, line 27: Does it currently fail for some of the vaults?

src/apps/sommelier/assets/sommelier.png
jeanregisser, line 1: Good for now, but we should add it to the dapp-list too.

src/apps/sommelier/positions.ts
```diff
@@ -0,0 +1,101 @@
...
+} from '../../types/positions'
+import { cellarV0821Abi } from './abis/cellar'
+import { getSommStrategiesData } from './sommelierApi'
+
+const hook: PositionsHook = {
+  getInfo() {
+    return {
+      id: 'sommelier',
+      name: 'Sommelier',
+      description: 'Sommelier vaults',
```
jeanregisser, line 19: They now market themselves as just "somm".
```suggestion
      id: 'somm',
      name: 'Somm',
      description: '', // Not sure anywhere
```

src/apps/sommelier/sommelierApi.ts
```diff
@@ -0,0 +1,86 @@
...
+        strategySlug: cellarAddressToSlug[cellarAddress.toLowerCase()], // used to generate the manageUrl
+      }
+    })
+}
+
+// Derived from the hard coded config in the Sommelier website source code
+// https://github.com/PeggyJV/sommelier-strangelove/blob/ca7bd6605bc868a1393d820f13b341ae5a5f1ead/src/utils/config.ts,
+// filtered by the vaults returned by
+// https://app.somm.finance/api/sommelier-api-all-strategies-data
+export const cellarAddressToSlug: Record<string, string> = {
```
jeanregisser, line 52: If they add a new vault we'll need to add it
```suggestion
export const cellarAddressToSlug: Record<string, string | undefined> = {
```

src/apps/sommelier/sommelierApi.ts
```diff
@@ -0,0 +1,86 @@
...
+    }>()
+
+  return result.data.cellars
+    .filter((cellar) => cellar.chain === NETWORK_ID_TO_SOMM_CHAIN[networkId])
+    .map((cellar) => {
+      const cellarAddress = cellar.id.split('-')[0] as Address // some cellars have a suffix to the address (e.g. `-arbitrum`), which we do not need
+      return {
+        ...cellar,
+        address: cellarAddress,
+        strategySlug: cellarAddressToSlug[cellarAddress.toLowerCase()], // used to generate the manageUrl
```
jeanregisser, line 43: Right now this will crash if they add a new vault.
What do you think of logging an warning if we don't have the slug?

src/apps/sommelier/positions.ts
```diff
@@ -0,0 +1,101 @@
...
+    const cellars = await getSommStrategiesData(networkId)
+    const positionDefinitions: PositionDefinition[] = []
+
+    const results = await Promise.allSettled(
+      cellars.map(async (cellar) => {
+        const [underlyingAsset, underlyingAssetSymbol, cellarName] =
+          await Promise.all([
+            client.readContract({
+              address: cellar.address,
+              abi: cellarV0821Abi,
```
jeanregisser, line 33: I noticed the config on their website uses different ABIs based on the vault.
Do they still implement the 3 methods called? 
Or would we need to tweak it per vault?

jeanregisser (approved):
Awesome work! 🚀 
Congrats on building your first position hook! 🥇

src/apps/sommelier/sommelierApi.ts
```diff
@@ -0,0 +1,86 @@
...
+      }
```
jeanregisser, line 44: What do you think of logging a wanring if we don't have the slug, so we can see we need to update the hardcoded list?

src/apps/sommelier/sommelierApi.ts
```diff
@@ -0,0 +1,86 @@
...
+    }>()
+
+  return result.data.cellars
+    .filter((cellar) => cellar.chain === NETWORK_ID_TO_SOMM_CHAIN[networkId])
+    .map((cellar) => {
+      const cellarAddress = cellar.id.split('-')[0] as Address // some cellars have a suffix to the address (e.g. `-arbitrum`), which we do not need
+      return {
+        ...cellar,
+        address: cellarAddress,
```
jeanregisser, line 42: nit: So we don't need to lowercase further down the line.
```suggestion
        address: cellarAddress.toLowerCase(),
```

src/apps/sommelier/positions.ts
```diff
@@ -0,0 +1,101 @@
...
+          tokens: [
+            {
+              address: underlyingAsset.toLowerCase(),
+              networkId,
+            },
+          ],
+          displayProps: () => {
+            return {
+              title: cellarName,
+              description: underlyingAssetSymbol,
```
jeanregisser, line 69: In a future improvement, it could be nice to include the APY in the description. 💡

docs/types/position.md
```diff
@@ -42,9 +42,17 @@ import { PositionsHook } from '../../types/positions'
 
 const hook: PositionsHook = {
   getInfo() {
-    return { name: 'MyDapp' }
+    return {
+      id: 'someId',
+      name: 'Some Dapp',
+      description: 'Some short description',
```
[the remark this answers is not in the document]
kathaypacific: ah, okay good to know! i will revert these docs changes

src/api/production.yaml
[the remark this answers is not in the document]
kathaypacific: yes! does staging use the testnet networks? i don't think that sommelier works on testnets but i notice that we've added other dapps to the staging config that i don't expect to work...what are we using the staging env for?

src/apps/sommelier/assets/sommelier.png
kathaypacific: yes i can do this in a follow up!

src/apps/sommelier/positions.ts
kathaypacific: this is confusing, i see that they are "somm" on their app but the documentation website is still "sommelier" 😅 i can update this whole thing to reference "somm" only to avoid confusion...

src/apps/sommelier/positions.ts
kathaypacific: yes i added a comment about this so that we remember, there are indeed 4 different cellar abi's but they all implement the methods that we need here. the main differences are in the events they emit and the non read-only methods, which we don't care much about here so i thought it was not worth also keeping a mapping of abis for our purposes

src/api/production.yaml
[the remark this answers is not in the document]
jeanregisser: All networks are enabled there too. 
Valora alfajores uses it, but it's not going to be around for very long.

src/apps/sommelier/positions.ts
jeanregisser: No that's good, it's better if it can still return what works.
I wonder if we should log an error instead of a warning, so we see something unexpected happened 🤔

src/apps/sommelier/positions.ts
jeanregisser: Yes, this is recent, it seems they'll gradually update everything.

src/apps/sommelier/positions.ts
jeanregisser: ahah yes I imagine it wasn't directly obvious.

src/apps/sommelier/positions.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const hook: PositionsHook = {
+  getInfo() {
+    return {
+      id: 'sommelier',
+      name: 'Sommelier',
+      description: 'Sommelier vaults',
+    }
+  },
+  async getPositionDefinitions({ networkId }) {
```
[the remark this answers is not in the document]
kathaypacific: ah, thanks for explaining! i was confused about how to use `address` before, since the actual balance was being added in the runtime i didn't think further on the topic. i added a check for the balance in the hook itself, but i agree at some point it might be nice for it to be in the runtime

src/apps/sommelier/positions.ts
kathaypacific: ah, i think i found the logic that they use to derive APY. i will add this as a follow up

jeanregisser (approved):
💯

PR #694, opened 2024-12-13
refactor: remove `id` and `description` from `AppInfo`

jeanregisser:
`description` is not used and the `id` is determined from the folder name. So removing to avoid confusion.

Addresses https://github.com/mobilestack-xyz/hooks/pull/692#discussion_r1883666876


kathaypacific (approved):
💅🏼

PR #695, opened 2024-12-13
refactor(i18n): share init between app and test code

jeanregisser:
We don't really need different behavior between app and test code.

Refactored to avoid possible confusion / things getting out of sync.

Also disabled i18next "debug" logs when using the `yarn getPositions` cli or running tests. It's not really needed and was cluttering the output IMHO.


finnian0826 (approved):
Looks good to me!

PR #704, opened 2024-12-16
feat(somm): add APY to position description

kathaypacific:
This PR adds some logic to calculate the "30D MA APY" (which I think means 30 day monthly averaged APY), based on Somm's website logic. The website logic is really quite involved and I really didn't want to copy over more logic than needed, so the function that I ended up with doesn't really look like the Somm source code but I've verified that the result exactly matches what is displayed on the website.

Related to RET-1258

### DIFF
--- a/src/apps/somm/api.ts
+++ b/src/apps/somm/api.ts
@@ -1,3 +1,4 @@
+import BigNumber from 'bignumber.js'
 import { Address } from 'viem'
 import { logger } from '../../log'
 import { NetworkId } from '../../types/networkId'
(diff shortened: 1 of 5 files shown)


src/apps/somm/api.ts
```diff
@@ -47,46 +117,220 @@ export async function getSommStrategiesData(networkId: NetworkId) {
...
+//
+// Strategy configuration reference:
+// https://github.com/PeggyJV/sommelier-strangelove/tree/ca7bd6605bc868a1393d820f13b341ae5a5f1ead/src/data/strategies
+const cellarConfig: Record<
+  string,
+  {
+    slug: string | undefined
+    estimatedApy?: number
+    deprecated: boolean
+    apyEnabled: boolean
```
jeanregisser, line 156: Why do we need to enable/disable this?

jeanregisser (approved):
amazing! 🚀

PR #712, opened 2024-12-23
fix: incorrect manage url for real yield eth vault

kathaypacific:
The manage url for the RYETH vault for Arbitrum was linking to the RYETH vault for Optimism (it's quite noticeable because the vaults have different APY and TVL's). The root cause is that I was using address for the cellar config key, which is not unique for all vaults. This PR updates the config to use the same id's as in the somm strategies api.

Related to RET-1258


src/apps/somm/positions.test.ts
```diff
@@ -148,13 +148,13 @@ describe('hook', () => {
...
                   {
-                    id: '0x392b1e6905bb8449d26af701cdea6ff47bf6e5a8-arb',
+                    id: '0x392b1e6905bb8449d26af701cdea6ff47bf6e5a8-arbitrum',
                     shareValue: '1050000',
                     tvlTotal: 10000000,
                     chain: 'arbitrum',
                   },
                   {
-                    id: '0xcellarAddressThatResultsInReadError-arb',
+                    id: '0xc47bb288178ea40bf520a91826a3dee9e0dbfa4c-arbitrum', // mockReadContract will fail on this address
```
bakoushin, line 157: supernit: what about instead of using a real chain name, use here a deliberately made-up one, like `unknown`, to emphasize it should fail?

bakoushin (approved):
LGTM! 🎉

src/apps/somm/positions.test.ts
kathaypacific: good idea, updated!

PR #720, opened 2024-12-30
feat(somm): add deposit shortcut

kathaypacific:
This PR adds the deposit shortcut for Somm cellars:
- add `dataProps` to show the cellars on the Earn screen. Note that this PR doesn't enable the Somm vaults for all apps yet, and some extra work will be needed to show info that the Earn screens currently don't cater for (I've made notes separately, examples are the platform/performance/exit fees, ability to deposit with multiple tokens)
- add positions shortcut. Note that this is pretty much copy paste from the Beefy shortcut, the logic is very much the same when depositing with the main cellar token (we will need to handle depositing with alternative tokens separately)

Related to RET-1258

https://github.com/user-attachments/assets/83902c48-4372-46b2-bd1e-04011171f3d8


src/apps/somm/positions.ts
```diff
@@ -46,26 +48,43 @@ const hook: PositionsHook = {
...
-              address: cellar.address,
-              abi: cellarV0821Abi,
-              functionName: 'name',
-            }),
-          ])
+        const [
+          underlyingAsset,
+          underlyingAssetSymbol,
+          cellarName,
+          cellarDecimals,
```
bakoushin, line 55: is this underlying asset decimals? or does the cellar have decimals as well?
sorry, i'm not well versed in how somm works, but it caught my eye

src/apps/somm/shortcuts.e2e.ts
```diff
@@ -0,0 +1,38 @@
...
+
+      const { transactions } = await shortcut!.onTrigger({
+        networkId: NetworkId['arbitrum-one'],
+        address: '0x2b8441ef13333ffa955c9ea5ab5b3692da95260d',
+        tokenAddress: '0xaf88d065e77c8cC2239327C5EDb3A432268e5831', // USDC
+        tokenDecimals: 6,
+        positionAddress: '0xa73b0b48e26e4b8b24cead149252cc275dee99a6', // RYUSD
+        tokens: [
+          {
+            tokenId: `${NetworkId['arbitrum-one']}:0xaf88d065e77c8cc2239327c5edb3a432268e5831`,
```
bakoushin, line 29: supernit: perhaps we can just use a plain string instead of template literal?

bakoushin (approved):
LGTM! ✨

src/apps/somm/positions.ts
kathaypacific: yes the cellar has decimals too 😅 i think that usually the cellar token has the same decimals as the underlying token, but i'm not sure that it is a necessity. since we use the cellar value for withdrawals, i thought it'd be more safe to use the cellar decimals directly

src/apps/somm/shortcuts.e2e.ts
kathaypacific: sure!

src/apps/somm/positions.ts
```diff
@@ -93,12 +112,44 @@ const hook: PositionsHook = {
...
+            termsUrl: SOMM_TERMS_URL,
+            manageUrl,
+            tvl: toSerializedDecimalNumber(cellar.tvlTotal),
+            yieldRates: [
+              ...(cellar.apy
+                ? [
+                    {
+                      percentage: cellar.apy,
+                      label: t('yieldRates.earningsApy'),
+                      tokenId: underlyingTokenId,
```
jeanregisser, line 128: The APY is the 30d average, right?
We may need/want to indicate this to the user. Maybe a custom string for this here.

jeanregisser (approved):
Looks great! 🚀

PR #721, opened 2024-12-30
feat(somm): add shortcut for withdraw

kathaypacific:
This PR adds the happy path for withdrawals. Somm cellars have a max withdraw amount, and a withdraw exceeding that amount needs to invoke a different withdraw flow which i'll add as a follow up.

Related to RET-1258

https://github.com/user-attachments/assets/ed24e10c-d403-4607-90b2-c4bb7b253539


src/apps/somm/shortcuts.e2e.ts
```diff
@@ -35,4 +35,30 @@ describe('getShortcutDefinitions', () => {
...
+      const shortcut = shortcuts.find((shortcut) => shortcut.id === 'withdraw')
+      expect(shortcut).toBeDefined()
+      expect(shortcuts[1].id).toBe('withdraw')
+
+      const { transactions } = await shortcut!.onTrigger({
+        networkId: NetworkId['arbitrum-one'],
+        address: '0x2b8441ef13333ffa955c9ea5ab5b3692da95260d',
+        tokens: [
+          {
+            tokenId: `${NetworkId['arbitrum-one']}:0xb9A27ba529634017b12e3cbbbFFb6dB7908a8C8B`,
```
bakoushin, line 53: as we agreed, this can be simplified to a simple string

src/apps/somm/shortcuts.e2e.ts
```diff
@@ -35,4 +35,30 @@ describe('getShortcutDefinitions', () => {
...
   })
+
+  describe('withdraw.onTrigger', () => {
+    it('should return transactions', async () => {
+      const shortcuts = await hook.getShortcutDefinitions(
+        NetworkId['arbitrum-one'],
+      )
+      const shortcut = shortcuts.find((shortcut) => shortcut.id === 'withdraw')
+      expect(shortcut).toBeDefined()
+      expect(shortcuts[1].id).toBe('withdraw')
```
bakoushin, line 46: do we need to check that `withdraw` is exactly 2nd?
looks like we checked above that at least one shortcut with an id = `withdraw` exists

bakoushin (approved):
LGTM!

jeanregisser (approved):
Awesome! 🚀

PR #726, opened 2025-01-02
fix(beefy): format pools pricePerShare correctly

satish-ravi:
https://valora-app.slack.com/archives/C04B61SJ6DS/p1735833604857669

### DIFF
--- a/src/apps/beefy/positions.ts
+++ b/src/apps/beefy/positions.ts
@@ -104,7 +104,13 @@ const beefyAppTokenDefinition = ({
         networkId,
       })
       const { decimals } = tokensByTokenId[tokenId]
-      return [toDecimalNumber(BigInt(vault.pricePerFullShare), decimals)]
+      return [
+        toDecimalNumber(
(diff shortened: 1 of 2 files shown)


silasbw (approved):

src/apps/beefy/positions.ts
```diff
@@ -104,7 +104,13 @@ const beefyAppTokenDefinition = ({
         networkId,
       })
       const { decimals } = tokensByTokenId[tokenId]
-      return [toDecimalNumber(BigInt(vault.pricePerFullShare), decimals)]
+      return [
+        toDecimalNumber(
+          // pricePerShare can potentially be a string in scientific notation, so convert to Number, round and then to BigInt
+          BigInt(Math.round(Number(vault.pricePerFullShare))),
```
jeanregisser, line 110: To avoid loosing precision:
```suggestion
          BigInt(new BigNumber(vault.pricePerFullShare).toFixed()),
```

Maybe we could also update `toDecimalNumber` to accept `string` as input.

PR #743, opened 2025-01-15
fix(swap-deposit): use the right networkId for swaps

satish-ravi:


MuckT (approved):
Thanks for fixing this! I tested locally with curl and a wallet instance with locally running hooks repo. I ran into a [revert](https://optimistic.etherscan.io/tx/0xc274b785aa66d605ee43596c89f5285c90a5e12fd03cbff203125c285012f4c3), but the swap was tx prepared.

PR #758, opened 2025-01-29
chore(scripts): triggerShortcut improvements

jh2oman:
setup for debugging https://linear.app/valora/issue/ACT-1546/fix-cross-chain-swap-deposits-from-erc20-tokens


scripts/triggerShortcut.ts
```diff
@@ -134,7 +134,9 @@ void (async () => {
   }
 
   const account = mnemonicToAccount(argv.mnemonic, {
-    path: argv.derivationPath as any,
+    ...(argv.networkId === NetworkId['celo-mainnet'] && {
+      path: argv.derivationPath as any,
+    }),
```
jeanregisser, line 139: I'd expect the derivation path to be used if passed from argv.
It seems what we'd want is change the default derivation path used (when not provided) when the network is Celo vs others. (so it behaves like our wallet).

scripts/triggerShortcut.ts
```diff
@@ -143,19 +145,24 @@ void (async () => {
     )
   }
 
+  const fromNetworkId: NetworkId = triggerInput?.swapFromToken?.networkId
```
jeanregisser, line 148: This seems a bit too specific.
Do we really need it?

jeanregisser (commented):
Thanks! 👍

PR #790, opened 2025-02-25
feat: optionally use allbridge header if provided

MuckT:
### Description

Add support for our Allbridge header to prevent rate-limiting; Users without this header / api-key should not be impacted.

### DIFF
--- a/src/apps/allbridge/api.ts
+++ b/src/apps/allbridge/api.ts
@@ -102,12 +102,9 @@ export async function getAllbridgeTokenInfo({
       'x-Sdk-Agent': 'AllbridgeCoreSDK/3.21.0', // as in https://github.com/allbridge-io/allbridge-core-js-sdk/blob/5deb0623d6fffcbc7a85dba22b98942c47d4af6c/src/client/core-api/api-client.ts#L48
     }
 
-    const allBridgeHeader = getConfig().ALLBRIDGE_HEADER
-    if (allBridgeHeader) {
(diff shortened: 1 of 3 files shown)


src/apps/allbridge/api.ts
```diff
@@ -97,12 +98,20 @@ export async function getAllbridgeTokenInfo({
     TOKEN_INFO_RESPONSE_KEY,
   ) as AllbridgeApiResponse
   if (!allbridgeTokensInfoResponse) {
+    const headers: Record<string, string> = {
+      'x-Sdk-Agent': 'AllbridgeCoreSDK/3.21.0', // as in https://github.com/allbridge-io/allbridge-core-js-sdk/blob/5deb0623d6fffcbc7a85dba22b98942c47d4af6c/src/client/core-api/api-client.ts#L48
+    }
+
+    const allBridgeHeader = getConfig().ALLBRIDGE_HEADER
```
jophish, line 105: Could we just specify the API key in the secret and assign it to a static header here? It seems like if something would change, it would be the API key value, and not the header key itself; this also removes room for error in setting up the secret (since nothing would complain about an incorrect header key in the secret itself).

jophish (approved):
Left an optional nit about setting up the secret, but A'ing as-is!

src/apps/allbridge/api.ts
MuckT: Changed in 5b41c91da28f015284c2747f819fd61e45ee799d!

PR #820, opened 2025-09-17
chore: cache `getTokensInfoWithPrices`

bakoushin:
### Description

Cache base tokens info for 5 seconds, while allowing stale-while-revalidate behavior.

This will ensure:
1. Data is always fresh. 5 second lag seems negligible, given the direct request may take 1-3 seconds
2. Data is retrieved fast

### DIFF
--- a/src/runtime/getPositions.ts
+++ b/src/runtime/getPositions.ts
@@ -33,15 +33,6 @@ import { TFunction } from 'i18next'
 import { getPositionId } from './getPositionId'
 import { LRUCache } from 'lru-cache'
 
-// Cache for getBaseTokensInfo with 10-minute TTL
-const tokensInfoCache = new LRUCache({
-  max: 1,
-  // In milliseconds
-  ttl: 10 * 60 * 1000,
-})
-
-const TOKENS_INFO_CACHE_KEY = 'base-tokens-info'
-
 interface RawTokenInfo {
   address?: string
   name: string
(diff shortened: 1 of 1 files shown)


src/runtime/getPositions.ts
```diff
@@ -54,23 +64,29 @@ export async function getBaseTokensInfo(
...
-  // Map to TokenInfo
-  const tokensInfo: TokensInfo = {}
-  for (const [tokenId, tokenInfo] of Object.entries(data)) {
-    tokensInfo[tokenId] = {
-      ...tokenInfo,
-      priceUsd: toSerializedDecimalNumber(tokenInfo.priceUsd ?? 0),
-      // We don't have this info here but it's not yet needed for base tokens anyway
-      balance: toDecimalNumber(0n, 0),
-      totalSupply: toDecimalNumber(0n, 0),
+  let tokensInfo = tokensInfoCache.get(TOKENS_INFO_CACHE_KEY) as TokensInfo
```
satish-ravi, line 67: The cache key should include network ids. Otherwise w/ different apps calling with different network ids, we'd return the wrong list of tokens.

Also, if we just want caching for the usdc dapp, we could pass an optional query parameter and only cache if its set. Would also solve the above problem as network ids from the dapp won't change

jeanregisser (commented):
What about setting needed cache headers at the endpoint producing these values?

The `got` client would respect that out of the box, If I remember correctly.

But if we want an in-memory cache, `got` also supports it.

satish-ravi (approved):

